PR #7, opened 2018-05-23
how about geometric shapes

fuddl:
how about adding [geometric shapes](https://en.wikipedia.org/wiki/Geometric_Shapes)?


fuddl:
Sorry, I will not be able to contribute here, since I do not own Adobe illustrator. 😐

b-g:
@fuddl I'm sorry too! Yes it is very lame for an "open source" project to rely on such proprietary tools, but currently we are simply very busy getting it to the next level. Hope fully we can base the next iteration more on an open SVG workflow. Sorry again!

dnlutz:
at the moment the only alternative for icon and emoji design is affinity designer:
https://affinity.serif.com/de/designer/
… or do you have any suggestions which software we could use?

loominade:
> … or do you have any suggestions which software we could use

if you'd switched on svg based workflow – as @b-g suggested – you could any modern vector graphic software to update your font, as well as a simple text editor.

carlinmack:
Work flow has been switched to SVG if anyone in this thread wants to contribute now :)

PR #36, opened 2019-10-04
Fix: Venezuela flag's color order

MahdiBaghbani:
Corrected Venezuela flag's color order .


b-g:
Hi @MahdiBaghbani, many thanks for this!! Just ... you fixed the file which is automatically generated. Could you apply the fix to the corresponding one in the `src/` folder and add your credits to the `changelog.txt`?

MahdiBaghbani:
someone else has created a better fix for this in #38, so I think it is better to close this one.

Issue #59, opened 2019-10-24
Some hexcodes in data/extras-unicode.csv are lowercase


sourgift:
For some reason three hexcodes in extras-unicode.csv are lowercase which prevents some of the helper scripts from running properly. 

I'm submitting a pull request to solve this issue.

PR #60, opened 2019-10-24
Incorrect hexcodes

sourgift:
This is to address #59 .  Some of the data has additional annotation as a result of running the generate-data-tables and generate-catalog scripts.


b-g:
@sourgift Thanks! 

> Some of the data has additional annotation as a result of running the generate-data-tables and generate-catalog scripts

Not on my end. Odd. Hence I've added your bugfix via a commit of mine https://github.com/hfg-gmuend/openmoji/commit/b6c4b5c7b5ca33c13cada71ef8426b13eec57a13. Hope you don't mind.

Issue #72, opened 2019-11-14
Improving consistency with other platforms - 12.1


carlinmack:
## Faces

- [x] Grimacing face isn't consistent - #61 
- [x]     Eyes should be to the right and not centered  [Smirking Face](https://emojipedia.org/smirking-face/) and [Unamused Face](https://emojipedia.org/unamused-face/)  - #74
- [x]   [face screaming in fear](https://emojipedia.org/face-screaming-in-fear/ ) Hands should be clasped around the face not towards the eyes #85 
- [x] [beaming face](https://emojipedia.org/grinning-face-with-smiling-eyes/) doesn't have bottom teeth showing - from issue #82 - #88 
- [x] [frowning face with open mouth](https://emojipedia.org/frowning-face-with-open-mouth/) - shouldn't have eyebrows - #100 
- [x] [tired](https://emojipedia.org/tired-face/) and [weary face](https://emojipedia.org/weary-face/) emoji consistently has a wide, down turned, open mouth — whereas our emojis have a closed mouth. - from #67 - #100 - 6fff9f07
- [x] If we are using big eyes like so [face with rolling eyes](https://emojipedia.org/face-with-rolling-eyes/) we should have similarly big eyes for [flushed face](https://emojipedia.org/flushed-face/) and [face with pleading eyes](https://emojipedia.org/face-with-pleading-eyes/ ) - #100 - f0d9f9f
- [ ]     [Shocked Face with Exploding Head](https://emojipedia.org/shocked-face-with-exploding-head/) looks like some sort of crown but should have a cloud of smoke emanating (we use this emoji as an example of line art on the contributions page)
- [ ] [face with look of triumph](https://emojipedia.org/face-with-look-of-triumph/) - complete redesign - There should be some steam coming out of nose 
- [ ]  [serious face with symbols covering mouth](https://emojipedia.org/serious-face-with-symbols-covering-mouth/) - redesign - There are no symbols covering the mouth and the emoji should potentially be red

### Face emojis with hands 
We have 5 fingers in these emoji [Face With Finger Covering Closed Lips](https://emojipedia.org/face-with-finger-covering-closed-lips/) and [Thinking Face](https://emojipedia.org/thinking-face/) and I think we should have 5 fingers in these emoji too
- [ ]     4 fingers - [smiling face with smiling eyes and hand covering mouth](https://emojipedia.org/smiling-face-with-smiling-eyes-and-hand-covering-mouth/)
[...]

PR #74, opened 2019-11-19
Consistency Issue #2 - moving eyes to the right

carlinmack:
Fixing issue 2 of #72


b-g:
Hi @carlinmack, many thanks for this! + Much appreciated! Just the files seem to be broken:

Screen Shot 2019-11-20 at 08 49 52
Screen Shot 2019-11-20 at 08 52 14

Side note: As they pass unfortunately the https://openmoji-tester.glitch.me/ app in "green" ... I will have to once more beef up the openmoji-tester app :(

carlinmack:
I noticed that the rich diff didn't show but I was confused why not, as they do display properly in inkscape if you copy the raw text. I'll submit another commit that removes the duplicate attribute.

If it's because of duplicate attributes it shouldn't be too hard to detect them, I might even have a go at implementing this.

carlinmack:
@b-g how can I run file-format.js locally?

b-g:
@carlinmack 

> How can I run file-format.js locally?

```
node_modules/.bin/mocha --grep "layer existing" --reporter mochawesome --reporter-options reportDir=test/report,reportFilename=report,json=false,code=false,cdn=true,reportTitle=OpenMoji-Tester,reportPageTitle=OpenMoji-Tester test/*.js --openmoji-data-json $PWD/data/openmoji.json --openmoji-src-folder $PWD/src
```

or better use the https://github.com/hfg-gmuend/openmoji-tester repo to develop tests ... there you can run a test on a single svg file

b-g:
@carlinmack 

Had a quick chat with @dnlutz on your contribution! Really better! Just we think that having the eyes that detailed is too much. We aim for minimalism and this would introduce another special case. Sorry! I hope you understand our point of view!

Could you?
- Eyes with circles only
- Add changes to changelog

Happy to merge it afterwards.

Thanks!!

carlinmack:
sure!

b-g:
@carlinmack Thanks! Merged!

PR #75, opened 2019-11-20
Add test for duplicate attributes

carlinmack:
Can't use the JSDOM version of the svg as it trims duplicate attributes, so I added a utility function to read the file to a string. Have to regex for the elements and then regex for the attributes. All existing emoji pass these tests

Test required due to #74


b-g:
@carlinmack Many thanks for it! Great! Merged most of it and added a more general is valid xml test by using https://github.com/libxmljs/libxmljs ... see the result in #77. 🙏!

PR #79, opened 2019-11-24
Fix melon colours

carlinmack:
#72


b-g:
@carlinmack Many thanks! Nice! But see github review for two small non melon related things. Afterwards happy to merge.

carlinmack:
Fixed contributions.txt, what else needs to be fixed? I added that line to beautify-svg.yml because we actually test for that

b-g:
@carlinmack Great! All good ... you are right `npm run pretty-src-svg`with convertStyleToAttrs` is even better. Merged.

Issue #31, opened 2019-09-24
Snowman is transparent


Pr0methean:
☃️ Snowman is transparent, whereas ⛄ Snowman Without Snow is opaque white, so they don't match when the background isn't white.
![Screenshot_20190923-170853](https://user-images.githubusercontent.com/4961925/65471708-045ef900-de25-11e9-9863-4be9bb863dcd.png)

dnlutz:
thanks for the hint. at the moment we have a some problems with openmoji on a dark background (to the left of the snowman there is a snowflake ; ) we have to think about how we can handle this.

b-g:
Many thanks for the bug report! I guess for the two snowmans it is easy to fix ... but as Daniel hinted there are quite a few cases like this, where we should find a general way of dealing with the background problem. You can either do a PR and fix the background or wait for the next official release, which aims to fix open issues and to introduce new special interest categories.

SlySven:
The simplistic solution for all emoji glyphs that have graphical elements that are thin and are monochrome is for a suitable contrasting colour outline around them so that it provides a background when the canvas onto which the glyph is painted is similar in colour as the original element...

![snowflake_lightbg](https://user-images.githubusercontent.com/6163092/67624879-ae6bdf80-f82e-11e9-89c8-36f652074f5a.png) ![snowflake_darkbg](https://user-images.githubusercontent.com/6163092/67624880-ae6bdf80-f82e-11e9-9ce3-2e301ef13fa8.png)

Pr0methean:
@SlySven Yes, this is what was done with Google's Noto Color Emoji to make them work on yellow backgrounds.

carlinmack:
not a fix to issue at large but fix to the snowmen in question in #128

dnlutz:
A 6px wide white contour line underneath would be the simplest and most pragmatic solution. It doesn't disturb on light backgrounds and can be removed if necessary. Actually I even like the extra line. The emojis look bolder and more stable on dark backgrounds.
<img width="947" alt="dark_mode" src="https://user-images.githubusercontent.com/38292227/82673196-c16a5500-9c41-11ea-84b8-f1ac0f0a3a15.png">

carlinmack:
I think this would generally work, but there's a few factors involved here:

1. The template
![image](https://user-images.githubusercontent.com/6676843/82701113-89611300-9c3d-11ea-8e00-26402d1a177c.png)
[...]

PR #86, opened 2019-11-29
Fix webbed fingers of ok and pinching emoji

carlinmack:
#72 The effect is visible on dark backgrounds 

image


b-g:
Many thanks! But I would like to discuss this with Daniel @dnlutz quickly. As it relates to one of the big open issues #31 ... the way forward might be simply a white additional outline around everything to differentiate our emojis properly on dark backgrounds.

carlinmack:
Trying to mock that up now and it's harder than it seems hmm

b-g:
@carlinmack Many thanks for this! But we will turn down this PR ... We think it is better to address and fix the issue in general and not just for a few ones. Sorry! Hope you understand.

carlinmack:
yep it's a bigger issue than this PR

PR #89, opened 2019-11-30
Fix foot to be more anatomically correct

carlinmack:
#72 

@dnlutz thoughts :)


carlinmack:
This is failing but not because of my changes 😶

dnlutz:
Sorry @carlinmack but I think it's hopeless to improve the old foot ; ) I would rather redraw it from scratch, attached a try.

I don't think we should even try to be anatomically correct. Instead, we should strive for maximum simplicity. Maybe not quite as blocky as Windows Emoji but as as geometric and icon-like as possible. 

geometric_foot

b-g:
@carlinmack Thanks! I agree with Daniel ... but as it is still an improvement I would leave it merged.
(PS: broken test ... yes not your fault. Was introduced earlier. We will fix it :)

PR #92, opened 2019-12-02
Fix emojis with speech bubbles to face correct direction

carlinmack:
#72 

image

Not failing due to my changes


b-g:
@carlinmack Will wait with all merges until we have the tests fixed. Sorry!

carlinmack:
One obvious problem with these right now is the lack of consistency in the way that the stem connects to the bubble. I think the spiky one has the most openmoji-consistent style but I didn't want to change the others without some discussion

b-g:
Many thanks for this! Yes good that the directions are now aligned to other vendors, also I find it more consistent/better how you re-designed the way stem and bubble are connected. Nice!

Just I don't think it is a good idea to introduce a half circle for 1F4AC. As we try hard to stay with minimal shapes and always aim to reduce complexity. Also it would be more consistent with the other bubbles.

OK?

2019-12-04 at 13 55

(Also there is a merged conflict on the changelog now)

carlinmack:
@b-g merging as I've changed from a semi circle to a full circle

PR #100, opened 2019-12-19
Create extra emojis and fix tired, frowning with open mouth faces

carlinmack:


carlinmack:
build is failing because it can't see the file?

b-g:
Hi @carlinmack Great! Just tests are red ... I believe it is simply because the file names are wrong.

Should be E280.svg E281.svg. Currently E280.svg 1E281.svg. Note the not correct `1` at the beginning of the file name.

carlinmack:
good catch, no idea how that happened

b-g:
@carlinmack Thanks for the quick fix! Merged!

carlinmack:
#72

PR #101, opened 2019-12-19
Create extra pomegranate emoji

carlinmack:


carlinmack:
I don't think this is good enough for two reasons:

1. probably too detailed
1. the non-colour version won't have any seeds. 

I tried a version with color+black line seeds but it looks quite strange with the 2px thickness.

b-g:
For my eye it doesn't look bad! :)

Yes the marked area with the lines coming together is not optimal
2019-12-19 at 12 09

> the non-colour version won't have any seeds.

Can be fixed easily with defining an additional layer `line-supplement` and then simply drawing the seeds in black see the example https://openmoji.org/styleguide/#special_layers

cc @dnlutz

carlinmack:
not important but the file accidentally sneaked in with PR #100, lets still improve it though :)

b-g:
@carlinmack Conflicts here as well. Would then vote to merge it after resolving the conflicts straight away ... and then improve it with commits direct to master (at least add the seeds in black in line-supplement).

carlinmack:
I think that should work lol, seeing as the file is already in the repo

carlinmack:
apparently not, I'll have time tomorrow to fix it

carlinmack:
@b-g merging as no conflicts now, I changed the lines like you showed in your picture

PR #104, opened 2019-12-20
add poppy emoji

carlinmack:


carlinmack:
@b-g did you want stem/leaf in the emoji?

PR #111, opened 2020-01-12
Fix Horse Racing emoji and add racing emoji

carlinmack:


b-g:
@carlinmack Thanks!! + Merged!

Issue #105, opened 2020-01-03
Improving consistency with other platforms 12.2


carlinmack:
_A continuation of #72_ 

## Faces
- [x] [confused face](https://emojipedia.org/confused-face/) No eyebrows, slightly down turned mouth - 8603bf6
- [x] [confounded face](https://emojipedia.org/confounded-face/) Crossed eyes and wavy mouth - #114 
- [ ]     [shocked face with exploding head](https://emojipedia.org/shocked-face-with-exploding-head/) looks like some sort of crown but should have a cloud of smoke emanating (we use this emoji as an example of line art on the contributions page)
- [ ] [face with look of triumph](https://emojipedia.org/face-with-look-of-triumph/) - complete redesign - There should be some steam coming out of nose 
- [ ]  [serious face with symbols covering mouth](https://emojipedia.org/serious-face-with-symbols-covering-mouth/) - redesign - There are no symbols covering the mouth and the emoji should potentially be red
- [ ] [loudly crying face](https://emojipedia.org/loudly-crying-face/) - eyes should be closed

## Other
- [x] [horse racing](https://emojipedia.org/horse-racing/)  The current emoji is suitable for "Horse Riding" but not so much racing, read [this tweet chain](https://twitter.com/jelenawoehr/status/1192242149510676481) for reference. I suggest making the back of the jockey flat and the feet higher up the horse - #111 
- [x] [mate](https://emojipedia.org/mate-drink/) I think that the top of this should be grey and the straw should have a grey stroke beneath the straw - 60742e4
- [x] [menorah](https://emojipedia.org/🕎/)  - [tweet](https://twitter.com/clapifyoulikeme/status/1233890658626260992) - Poor, should be 9 candles - e28d774
- [x] [dotted six pointed star](https://emojipedia.org/dotted-six-pointed-star/) and [star of david](https://emojipedia.org/star-of-david/)- the star should be the star of david which is explicitly two connected triangles on top of each other - 414fe7d
- [x] [falafel](https://emojipedia.org/🧆/) - [tweet](https://twitter.com/VinnieTesla/status/1188903133230981120) - other platforms have a plate of falafel - 779a371 by @kraasch 
- [ ] [pinching fingers](https://emojipedia.org/pinching-hand/)  Pinching fingers should be closer
- [ ] [oyster](https://emojipedia.org/oyster/) - redesign - I don't think this should be purple at all
[...]

PR #114, opened 2020-01-20
Fix confounded face to be more consistent

carlinmack:
#105 

https://emojipedia.org/confounded-face/

think this works, but can be improved


b-g:
@carlinmack Many thanks + merged! :)

Issue #113, opened 2020-01-20
Science fiction rocket ship


kriskowal:
At time of writing, the rocket ship emoji is represented by the American space shuttle. Many other emoji fonts use a depiction of an iconic 60's era science fiction rocket ship.

https://emojipedia.org/rocket/

I propose a depiction more in keeping with the prevailing trend.

Since I'll be needing this regardless for a game I'm writing (a roguelike about and using emoji), I've begun drafting a prototype, attached.

![1F680](https://user-images.githubusercontent.com/60294/72709215-4ec55d80-3b19-11ea-813c-35dfd3fc4fb5.png)

b-g:
@kriskowal Many thanks for this + looks nice! 

I would propose to move the Space Shuttle to private use area and replace rocket 1F680 with it. 

Personally I would add to the discussion that I find a vertical rocket better and a science fiction rocket ship would always have some functional marks on the outer shell e.g. see Tin Tin Rocket Cover.

![220px-The_Adventures_of_Tintin_-_16_-_Destination_Moon](https://user-images.githubusercontent.com/480224/72710453-6cb9af00-3b67-11ea-804c-cbd43b95c19b.jpg)
 
cc @carlinmack is the master of consistency with other vendors :) What do you think?

carlinmack:
+1 for science fiction rocket ship and move NASA space ship to openmoji-extra, love the design! Think it would be fun to test a checkboard colouring like above

kriskowal:
![1F680-drafts](https://user-images.githubusercontent.com/60294/72750491-d8a11500-3b71-11ea-9a50-88adbbc85620.png)

kriskowal:
I believe I'll be needing to reserve a number in the private use area, or stack my PR over a separate change that moves the space shuttle into the openmoji-extras before me. Please advise. I still have a bit of an adventure ahead of me, figuring out how to tease Inkscape into exporting this with XML attributes instead of CSS styles.

carlinmack:
I need to write a proper guide how to do it, but if you `File -> Save As -> Save as type: -> Optimised SVG -> Save` and use these settings, you get pretty close to clean XML

![image](https://user-images.githubusercontent.com/6676843/72760224-1324a500-3ba6-11ea-911e-6a6cc8a8daeb.png)

carlinmack:
[...]

PR #117, opened 2020-01-21
Move space shuttle to private use area as E20B

kriskowal:
In preparation for #113, this change copies the space shuttle to the private use area for travel-places.


kriskowal:
I've run the scripts to generate tables, but there is a great deal of churn that I cannot explain. Let me know if there's anything I can do to make this diff more quiet.

carlinmack:
it might be due to one of the other people on this repo introducing the androgynous emoji's and not generating the tables for it. I'll check the diff more closely in the morning, looks fine though!

PR #128, opened 2020-02-12
Standardise snowmen

carlinmack:
Wanna make sure that I've made the right choice here


b-g:
@carlinmack Many thanks + Nice! Merged!

 I think this is a pragmatic fix of the snowmen! 

The arms and hat still won't really work on a dark background ... but this has to be discussed and fixed in general terms. I'm hoping that we might be able to recycle some knowledge we hopefully will gain by fixing the colorfont e.g. convex hull with paper.js to automatically generate a white background "behind" all existing OpenMojis ...

PR #137, opened 2020-03-07
New design for falafels.

kraasch:
Plate now has white background, file normalized and named accordingly.


carlinmack:
hey I've fixed the error and made some adjustments to your emoji, I tried to be as detailed as possible in the commits so that you can see whats going on.

you can see in https://github.com/kraasch/openmoji/compare/9861187...e04bac0?short_path=5f24453#diff-5f24453f52fc4e329ee60cb0174696a4 that I didn't really change much

What I did change was:

1. separating the fills to be on a separate layer - this is important so that we can easily generate line versions of the emoji
2. changing the line endings where different items intersect to be rounded
3. centring the felafel on the plate

b-g:
@kraasch @carlinmack Many thanks for the yummy new falafels! 😋

PR #141, opened 2020-03-19
fix rest of square purple religion and zodiac emojis

carlinmack:


b-g:
@carlinmack Super! Many thanks! Merged!

PR #151, opened 2020-03-29
Scripts review

kriskowal:
This change is a review of all the helper scripts to apply some bash best practices and get them passing the shellcheck linter. This change adds a github CI step to verify that the scripts pass shellcheck, and a package script for running shellcheck manually, though this requires `shellcheck` installed on the system.

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,8 +17,9 @@
     "colorfont"
   ],
   "scripts": {
-    "test": "node_modules/.bin/mocha --reporter mochawesome --reporter-options reportDir=test/report,reportFilename=report,json=false,code=false,cdn=true,reportTitle=OpenMoji-Tester,reportPageTitle=OpenMoji-Tester test/*.js --openmoji-data-json $PWD/data/openmoji.json --openmoji-src-folder $PWD/src",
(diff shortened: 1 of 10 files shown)


package.json
```diff
@@ -17,8 +17,9 @@
     "colorfont"
   ],
   "scripts": {
-    "test": "node_modules/.bin/mocha --reporter mochawesome --reporter-options reportDir=test/report,reportFilename=report,json=false,code=false,cdn=true,reportTitle=OpenMoji-Tester,reportPageTitle=OpenMoji-Tester test/*.js --openmoji-data-json $PWD/data/openmoji.json --openmoji-src-folder $PWD/src",
-    "test-dev": "node_modules/.bin/mocha --invert --grep \"production\" test/*.js --openmoji-data-json $PWD/data/openmoji.json --openmoji-src-folder $PWD/src",
```
kriskowal: The relevant changes here are the removal of `node_modules/.bin` since `npm` ensures this is on the `PATH` and the removal of `$PWD/` since this is also ensured to be the project root by npm.

kriskowal: Odd. Had to put the PWDs back. Must be something odd about mocha specifically.

PR #152, opened 2020-03-29
Improve export-png performance

kriskowal:
This change parallelizes PNG generation to reduce build times.


kriskowal:
On my machine, this reduces export-png latency from 5m to 2m.

kriskowal:
Oh, note that master color and black are stale. I verified that this change produces the same results by running the old version, committing a baseline, and running the new version.

kriskowal:
This change is subsumed in #156. I recommend revisiting this change only if #156 proves too complex to maintain.

PR #153, opened 2020-03-30
Add support for exporting files for one hexcode

carlinmack:
The first argument to the all the files I modified can now be the hexcode of the symbol. If no argument is passed then it runs as it normally does. I feel like this is a worthwhile change to speed up exporting time


carlinmack:
the github diff isn't very good, but basically all the deleted lines are in the `else` statement of each file

carlinmack:
I'll add documentation if you merge this :)

PR #149, opened 2020-03-28
Speed up pretty svg with a memo table

kriskowal:
This change introduces a helper that lets `pretty-src-svg` use a git note to avoid rebuilding the same output for the same input on every run.  On my machine, this reduces the build time from minutes to seconds for incremental updates.  This would result in a similar speed-up between CI and CD runs where the git caches are preserved between runs.

This change depends on a fix to svgo https://github.com/svg/svgo/pull/1237 which will need to be released before landing this change, and for an update to package.json and package-lock.json to use the patch.

Issue #154, opened 2020-03-30
Better build system


b-g:
Hi @kriskowal and @carlinmack,

Wow!! Many thanks for the PR!! As they are all build system related (and somehow conflicting each other) I thought we should bundle the discussion here. I hope you don't mind.

Comments:

@kriskowal 
- #149 this is a draft and if currently not intended for being merged, or? I guess we won't be able to merge it until your bug fix for SVGO was accepted. 
- #151 and #152 is ready to be merged, correct? 

@carlinmack 
- #153 Many thanks ... but this seems to be already solved with the PRs above.

Hmm ... I have to say I extremely appreciate the PRs, but I'm a bit overwhelmed by them. I think we (Daniel, Hartmut and me) are less experts in the field of dev ops and it is quite difficult for us to evaluate the implications of accepting them. After all I will have to maintain OpenMoji with code which I currently don't fully understand. This will create kind of a new dependency to @kriskowal :)

Hence may I suggest the following:

1.
- we create a branch `dev-build-system`; add you both as contributors
- we merge #151 #152 to branch `dev-build-system`
- we all test it
- iterate on it together
- once we are all fine with -> merge it with master

2.
Afterwards ... same process for #149 (if SVGO is patched) **or** we add a temporary feature e.g. `export single hexcode` as propsed in #153 

Is this an approach the two of you could support/accept?

carlinmack:
I'm not sure that @kriskowal s PR will solve my problem as I seem to generate different files that are visually identical with the current export scripts. I think the caching of input technique will maybe help this though.

kriskowal:
#149 is an earnest proposal to speed up incremental pretty-src-svg and does indeed require an svgo bugfix before it’s ready for review.

#151 is also an earnest proposal, but loaded with bash style guide opinions worth review. Since it touches all the scripts, it’s likely to cause conflicts. It may be better as a reference if you continue to evolve other scripts. I do think it will be helpful to add shellcheck to CI before larger sweeping changes since it establishes a good baseline for style.

#152 is merge-able but I believe @carlinmack and I are both continuing to experiment on further improvement. I agree being able to pass codes as variadic arguments to focus the build is useful. I also have an experiment open to add caching. To get good performance from caching, I’m taking a more novel approach than I did with pretty-svg-src memo-in-place. It’s possible that this approach will be fast enough for incremental builds that it provides a better developer experience than specifying codes as arguments.

[...]

PR #156, opened 2020-03-30
Introduce cached parallel build utility

kriskowal:
This change introduces a utility that serves a similar role as `xargs -P` in that it runs a build step with parallel subprocesses, but also remembers the input and output hashes from the prior run to avoid duplicate work. Every run benefits from parallelism. Subsequent runs benefit from the cache.

The solution batches the hash checks before and after to benefit from fast bulk git operations.

Separate commits apply this tool for `export-png` and `pretty-src-svg`. These could conceivably be used for `export-svg` after further refactoring. At that point, incremental builds would take seconds.

This is an alternative approach to #149, #152, and maybe #153.


kriskowal:
I rewrote the jq script in js, so there’s no longer a dependency.

kriskowal:
I am putting this project on pause. I may revisit this later, with a fresh wind. For now, please feel free to close.

If I revisit this later, my planned changes are hinted on #154.

1. Either thread a list of hexcodes through from generate.sh into all of the utility helpers or decouple bulk release generation from incremental generation entirely.
1. Use HEAD as the cache instead of a side table. It occurs to me that the prior committed state is a sufficient source of reusable hashes, as long as they remain checked in.
1. Consider shooting for a demonstration of CI/CD with asset generation.
1. Consider using a Docker container in CI/CD that enforces consistent versions of system dependencies, rsvg-convert and pngquant.

b-g:
Great!! But let's discuss in a quick call how to proceed. See https://github.com/hfg-gmuend/openmoji/issues/154#issuecomment-609610984

b-g:
Hi @kriskowal, I've spent some time with your proposed PR ... I think I understand the basics now ... WOW! Holy moly this is a lot faster! This is exactly what we need for the project. Many many thanks!!

Comments:

- Could we move every script which is supposed to be not directly called by the user to something like e.g. `/helpers/lib/`? I would find it a lot easier mentally that everything on the level of `/helpers/*` is a "batch script".
- Not familiar with `git update-ref refs/memos/` ... How can I see the memos? How can I commit them to the Github repo? (It seems this info not showing up in the GUI of my Github Desktop App :) 
- Could you create a skeleton "how to talk to" `optimize-build.sh` from node.js? Then I could refactor the node script inside `export-svg.sh`

kriskowal:
For sure. I like the idea of moving the helper helpers down a level.

`update-ref` is the mechanism underlying named branches and tags. Local branch names are shorthand for `refs/heads/*` and tags are shorthand for `refs/tags/*`. Remote branches are under `refs/remotes/$REMOTE/$BRANCH`. So, `git push/pull "$REMOTE" "'$BRANCH"` is actually shorthand for `git push/pull "$REMOTE" "refs/heads/$BRANCH:refs/heads/$BRANCH"` and has the side-effect of fixing `refs/remotes/$REMOTE/$BRANCH` to point to the same hash, all using `update-ref`.

While it is possible to push and pull memos using the fully qualified notation, `git push origin refs/memos/*`, it should never be necessary. Each local repository may have its own cache. I’ve used this memos namespace specifically to hide it from tools.

I’m not sure how to factor the export-svg scripts. At a high level, it’s far easier to call git from bash than it is to call git from node, so I try to arrange the problem around the former. It may be necessary to divide the export-svg scripts into a pipeline with separate plan | optimize | run stages. This would involve dividing each export-svg script into a plan script and run script, possibly both being in node, then orchestrate these two from a shell script. The plan script would need to generate a file with entries like:

```
color/svg/XXX.svg:src/A/B/C.svg
color/svg/YYY.svg:src/A/B/D.svg
```

The column order is not important to the optimizer, so you write this in whatever form you want to read back in the execute phase. The optimizer is just filtering out the entries in the plan for which the corresponding hashes for all of the files in the plan are the same as they were after the previous run.

The speed of this system lies in pipelining. This avoid the overhead of running shell loops and the overhead of starting a node process and loading all its dependencies. So the execute phase would want to loop over the entries in the shard of the plan it receives and run until all of them have been processed.

b-g:
Hi @kriskowal, many thanks for the explanation! It start to make more and more sense :)

> Each local repository may have its own cache. I’ve used this memos namespace specifically to hide it from tools.

OK. Sounds good! I assume that this would then work even on the first run on a new local machine because of diffing to the most up to date git hashes, correct?

> RE: refactoring export-svg scripts

```
color/svg/XXX.svg:src/A/B/C.svg
color/svg/YYY.svg:src/A/B/D.svg
```

This seems to assume a 1:1 relation between src svg file and deduced assets.
But this is not the case. Because of skintones and multi skintones a single e.g. https://openmoji.org/library/#search=hand&emoji=1F46C can have a lot of decedents https://openmoji.org/library/#search=hand&emoji=1F468-1F3FF-200D-1F91D-200D-1F468-1F3FC ... 

But if I had the information which src svg had been changed, then I would be enough to at least just export the needed ones. What about something really simple like? Yes then export-svg.js would not have a memo mechanism ... but it still would be extremely faster!

```generate.sh (pseudo code)
node helpers/generate-data-tables.js
node helpers/generate-catalog.js
CHANGED_SRC_FILES = optimize-build.sh ... or some kind of getChangedSrcSvgFiles()
bash helpers/pretty-src-svg.sh
#export-svg.js as a place holder for the future refactored script
node helpers/export-svg.js $CHANGED_SRC_FILES
npm run export-svg-font
bash helpers/export-png.sh
```

Once I have something like `node helpers/export-svg.js $CHANGED_SRC_FILES` I could do that part. What do you think?

kriskowal:
The plan optimizer does not actually assume a one to one source to target mapping. You can use any number of colon delimited files. The number of files does not need to be consistent from line to line. By convention, I believe it makes the most sense for the first entry to be the target file, and all remaining be source file names. The run phase does not need to even use all of these file names directly. They can be there just to inform the planner that the build must run if any of them have changed.

The optimizer will work even if the memo is absent, but it will generate everything the first time then write an initial hash memo. As long as stuff is checked in, and we trust what was checked in to be correct, we could have the tool use the master branch as the source of truth when the memo is absent, to speed up the first run.

kriskowal:
Regarding trust, the tool accepts a NO_MEMO=on environment variable to force regeneration of everything. I think this should be set when doing a final run for releases.

kriskowal:
I’ve pushed changes. Now the entire pipeline runs shards in parallel and caches the results, including export-svg. I’ve added a find-emojis.js helper, which helps building plans for each of the stages, but can also be used to limit the scope of a build to specific hex codes or other criteria, like `node helpers/find-emojis.js --has-skintones`.

I’ve also moved the scripts the user should not interact with directly into helpers/lib. Please let me know if I’ve missed anything.

A cold build on my machine takes 5 minutes. With a hot cache, it takes 15 seconds.

b-g:
@kriskowal Many thanks! This seems to be ready to go! Yay! This is soooo much faster now!

@carlinmack could you also test it e.g. `npm run generate` and double check that it is not just running fine on Chris' and my machine?

kriskowal:
I suggest running `time npm run generate`, twice :-)

carlinmack:
I'll do this tonight, trying to sort out my university work and deadlines right now :) glad to see that this is almost complete!

carlinmack:
run one was successful, two not so much
```
________________________________________________________
Executed in  120.01 mins   fish           external
   usr time   88.60 mins    2.85 millis   88.60 mins
   sys time   71.97 mins    4.60 millis   71.97 mins

carlin@SB2v2 /c/U/c/D/openmoji (export-png-cache)>
time npm run generate

> openmoji@12.2.0 generate /c/Users/carli/Desktop/openmoji       
> ./helpers/generate.sh

-------------------------------------------
👉 helpers/generate-data-tables.js
-------------------------------------------
👉 helpers/generate-catalog.js
-------------------------------------------
👉 helpers/pretty-src-svg.sh
warning: CRLF will be replaced by LF in helpers/illustrator/README.md.
The file will have its original line endings in your working directory
No changes
-------------------------------------------
👉 helpers/export-svg.sh
Exporting skintone SVGs
warning: CRLF will be replaced by LF in helpers/illustrator/README.md.
The file will have its original line endings in your working directory
1440 changes
Error: Error: no such file or directory 'color/svg/1F44B-1F3FD.svg'.
Error: Error: no such file or directory 'color/svg/1F44B-1F3FB.svg'.
Error: Error: no such file or directory 'color/svg/1F44B-1F3FE.svg'.
...
```

b-g:
@carlinmack cc @kriskowal 

Thanks! Hmm ... 

- Can you share the spec of your environment?
- Can you run the existing helpers on your environment without issues?

- > run one was successful, two not so much

What do you mean with hits exactly? Did you run `npm run generate` twice, or different commands (`time npm run generate` and then `npm run generate`)?

kriskowal:
There are hints in the output that this might be Windows Subsystem for Linux. I might be able to reproduce on a Windows machine. I’ve seen the missing file errors during development, but only if the prior phase failed to create them.

Did the first run truly take two hours?

carlinmack:
I ran `time npm run generate` twice. I'm running WSL2 on a preview build of Windows, however things like this are normally stable

carlinmack:
I ran it again this morning after checking out the branch again

```
carlin@SB2v2:/c/Users/carli/Desktop/openmoji$ time npm run generate

> openmoji@12.2.0 generate /c/Users/carli/Desktop/openmoji
> ./helpers/generate.sh

-------------------------------------------
👉 helpers/generate-data-tables.js
-------------------------------------------
👉 helpers/generate-catalog.js
-------------------------------------------
👉 helpers/pretty-src-svg.sh
warning: CRLF will be replaced by LF in helpers/illustrator/README.md.
The file will have its original line endings in your working directory
6 changes
Caching refs/memos/pretty-src-svg
-------------------------------------------
👉 helpers/export-svg.sh
Exporting skintone SVGs
warning: CRLF will be replaced by LF in helpers/illustrator/README.md.
The file will have its original line endings in your working directory
No changes
Exporting non-skintone color SVGs
warning: CRLF will be replaced by LF in helpers/illustrator/README.md.
The file will have its original line endings in your working directory
No changes
Extracting black outline SVGs from all color SVGs
warning: CRLF will be replaced by LF in helpers/illustrator/README.md.
The file will have its original line endings in your working directory
No changes
-------------------------------------------
👉 helpers/export-png.sh
Scale 72x72
warning: CRLF will be replaced by LF in helpers/illustrator/README.md.
The file will have its original line endings in your working directory
6956 changes
Caching refs/memos/export-png-72
Scale 618x618
warning: CRLF will be replaced by LF in helpers/illustrator/README.md.
The file will have its original line endings in your working directory
6956 changes
Caching refs/memos/export-png-618
-------------------------------------------
👉 helpers/clean.sh

real    42m23.118s
user    43m31.150s
sys     8m27.331s
```

carlinmack:
subsequent runs are ~15m :)

the change is that I was using `fish` in the original message rather than `bash`

kriskowal:
I wonder what is different when running bash scripts under fish.

carlinmack:
I'm trying again with fish, will report back

carlinmack:
damn for whatever reason fish is much slower, not a problem though, will just remember to use bash:
```bash
carlin@SB2v2 /c/U/c/D/openmoji (export-png-cache)>
time npm run generate

> openmoji@12.2.0 generate /c/Users/carli/Desktop/openmoji
> ./helpers/generate.sh

-------------------------------------------
👉 helpers/generate-data-tables.js
-------------------------------------------
👉 helpers/generate-catalog.js
-------------------------------------------
👉 helpers/pretty-src-svg.sh
No changes
-------------------------------------------
👉 helpers/export-svg.sh
Exporting skintone SVGs
No changes
Exporting non-skintone color SVGs
No changes
Extracting black outline SVGs from all color SVGs
No changes
-------------------------------------------
👉 helpers/export-png.sh
Scale 72x72
No changes
Scale 618x618
No changes
-------------------------------------------
👉 helpers/clean.sh

________________________________________________________
Executed in   25.55 mins   fish           external
   usr time   50.29 secs  1263.00 micros   50.29 secs
   sys time  215.64 secs  483.00 micros  215.64 secs

carlin@SB2v2 /c/U/c/D/openmoji (export-png-cache)>
time npm run generate

> openmoji@12.2.0 generate /c/Users/carli/Desktop/openmoji
> ./helpers/generate.sh

-------------------------------------------
👉 helpers/generate-data-tables.js
-------------------------------------------
👉 helpers/generate-catalog.js
-------------------------------------------
👉 helpers/pretty-src-svg.sh
No changes
-------------------------------------------
👉 helpers/export-svg.sh
Exporting skintone SVGs
No changes
Exporting non-skintone color SVGs
No changes
Extracting black outline SVGs from all color SVGs
No changes
-------------------------------------------
👉 helpers/export-png.sh
Scale 72x72
No changes
Scale 618x618
No changes
-------------------------------------------
👉 helpers/clean.sh

________________________________________________________
Executed in   37.83 mins   fish           external
   usr time   73.49 secs    0.00 millis   73.49 secs
   sys time  317.05 secs    6.80 millis  317.04 secs
```

b-g:
Many thanks!

@kriskowal I've spent 1h trying to merge your PR to the branch `buildsystem-with-cache` ... without luck :( I guess I'm not Git Pro enough for this / the Github Desktop App struggles to deal with this. I've added you as contributor with write access to the repo (please don't push things to Master without notice). Could you bring your PR to the branch? I would like to bring the font generator workflow up to speed and add some documentation (helpers/README.md) ... and hopefully merge then everything next week.

@carlinmack Could you add then to helpers/README.md potential Windows specific hints?

kriskowal:
There were a few merge conflicts. I’ve pushed my updates to both my branch
and yours, so the pull request should be merge-able, but if you would
like @carlinmack to add notes, best use your branch. Thank you!

carlinmack:
I don't really have anything specific to add to the readme. It's very unlikely that anyone would be using fish.

carlinmack:
Just to confirm, this can be merged as is, I think @b-g

Issue #161, opened 2020-04-03
Cetacean Proposal


kriskowal:
I have a proposal for additional cetaceans.

I propose the existing whale emoji be moved to the supplemental set as "orca", "black fish" or "killer whale".

To be more consistent #143 with other platforms, I propose a new blue whale. The basis is a 180 degree turn of the orca.

I've added variants for "narwhal" and "beluga" for sake of completeness (the narwhal is for my emoji quest game project).

![cetacean-plate](https://user-images.githubusercontent.com/60294/78317524-593ba480-7517-11ea-927a-b8f24ddfb5c0.png)

kriskowal:
Added cuts for 2.5D effect on whale flippers. I'm recommending whale and spouting whale variants.

https://emojis.wiki/whale/
https://emojis.wiki/spouting-whale/

![whales](https://user-images.githubusercontent.com/60294/78319379-e7198e80-751b-11ea-83ef-bc1b86968d56.png)

carlinmack:
wow! I love that we have a complete underwater mammal set

~~I don't really get rotating the orca, it's hard for me to read from the silhouette due to the fin being on the bottom, meaning it's upside-down. ~~ I pass to @b-g about whether orca or blue whale is our Whale design

I would be happy to merge this when the following design issues are resolved:
* adjust the shape of the narwhals head to be more accurate to the shape of actual narwhal heads https://commons.wikimedia.org/wiki/Category:Monodon_monoceros_(illustrations) 
* i didn't read properly and thought it was meant to be a dolphin, here's a lovely photo I was going to ask you to use lol https://commons.wikimedia.org/wiki/File:2017-05-16_Show_in_Kyiv_Dolphinarium_11.jpg
* your beluga seems to have broken it's back :( You can exaggerate from this within reason, but the main takeaway is more bending in the tail than the head https://commons.wikimedia.org/wiki/File:Delphinapterus_leucas_in_shallows.jpg

kriskowal:
The rotated orca is not a proposal. I used the shape as a base reference to
maximize consistency of the vocabulary. I think I will try the same flipped
along the vertical axis to produce a consistent porpoise 🐬

kriskowal:
[...]

PR #167, opened 2020-04-05
Add narwhal, beluga, porpoise, and replace dolphin

kriskowal:
This change adds a consistent set of cetaceans, building on the existing orca
and blue whale (spouting and otherwise).

This replaces the existing dolphin.
The others are new extras.

Per cetacean proposal #161


b-g:
Dolphin: 

- I think the grey of the "belly" should be under the fin, see the emojis of other vendors
2020-04-06 at 12 01
- also it might look better if the top fin is slightly more downwards (very rough sketch not a suggestion for the final curve) e.g.
2020-04-06 at 12 01

b-g:
- Find the angle of the horn slightly of the general direction / momentum of the body
- Would not add a grey fill on top of the horn
2020-04-06 at 12 11

b-g:
Very nice! Tiny detail: the line endings / curve join points of the light grey fin could be probably improved. The position where the fin joins the body looks too small IMO.

2020-04-06 at 12 14

b-g:
Same issue like described above at the dolphin. 

2020-04-06 at 12 18

b-g:
@kriskowal But really nice set! 🙏

kriskowal:
Draft, before recutting.

extra-cetaceans-plate

Also, I've just cut out new codes in the private use area. I'm open to combining, say, whale + party horn = narwhal, whale + melon = beluga, dolphin + hippopotamus = porpoise. I'm sure whatever reasoning for combining will be inscrutable, but, your option.

carlinmack:
We normally just cut out space in the private use area and name them E400, E401, E402 etc, but I like your style haha

kriskowal:
I'll leave them in the private use area. Seems wisest in any case. I'll wait for another round of feedback from @b-g before updating the change.

kriskowal:
Ah, and please weigh in on the color of dolphins. The other emoji sets are split between grey and blue, but more are blue. The narwhal and beluga should definitely be grey. I like the idea of distinguishing dolphin and porpoise in ways more obvious than the nose and fin, which is why I elected to flip along the vertical axis. Differentiating by color would I think be good for contrast. As always, I defer to your opinion!

b-g:
@kriskowal Wow! That's all great (Row 2)! Happy to merge!

>  color of dolphins?

This is a hard one :) I would rather go with grey, simply as our other emojis in animal-marine have rather realistic colors and not so much the "comic-esk" blue. But please decide on your own, also if you think it helps to differentiate between porpoise blue is fine too. 

> Also, I've just cut out new codes in the private use area. I'm open to combining, say, whale + party horn = narwhal, whale + melon = beluga, dolphin + hippopotamus = porpoise. I'm sure whatever reasoning for combining will be inscrutable, but, your option.

Nice idea! But I would rather go with a single private area usecode to not increase the complexity further. Ideally a future Unicode Emoji Version has all those animals ... and then we can move them to standard Unicode hex code.

b-g:
@kriskowal will wait for you final dolphin color decision and then merge. Let me know.

kriskowal:
This is once again in a reviewable state.

I don’t know whether everyone already found this, but for my own verification, I found the button that lets you look at the images in the diff.

Screen Shot 2020-04-08 at 4 42 46 PM

b-g:
@kriskowal Super! + Merged!

PR #171, opened 2020-04-11
Make orca fins consistent with other cetaceans

kriskowal:
This revises both orca and spouting orca to use the same fin as other cetaceans. Since this is not a substantial change to the original by Selina Bauer, it is my preference to leave the attribution alone.


b-g:
@kriskowal Thank you! Merged!

PR #174, opened 2020-04-17
Hexagon and oval

loominade:
I tried the add the [Black](https://unicode-table.com/en/2B22/) & [White Hexagon](https://unicode-table.com/en/2B21/) and [Black](https://unicode-table.com/en/2B2E/) & [White Vertical Ellipse](https://unicode-table.com/en/2B2F/).

Please let me know if there is anything wrong with it


carlinmack:
Thank you!

As these aren't official emojis, but unicode characters, you should add them to `src\extras-unicode\symbol-other` and `data\extras-unicode.csv` instead of `src/symbols/geometric/` and `data/enhancements-emoji-unicode-data.csv`

b-g:
I'm already on it ... should be fixed any minute! :)

@loominade Many thanks for the contribution!

PR #179, opened 2020-04-18
tag isometric emoji

fuddl:
did I miss any?


carlinmack:
the new mouse trap emoji is ortho too (it's in a recent commit), let me also check the openmoji library to find some more

carlinmack:
ice too. I think that's all :)

fuddl:
roll of paper maybe?

carlinmack:
yeah I would say so

carlinmack:
please add the tag to Ice too

carlinmack:
I didn't realise this but we're actually referring to isometric! 

image

can you do a ctrl+h (find and replace) for orthographic -> isometric then I'll merge haha

fuddl:
should totally be fixed in the [styleguide as well](https://openmoji.org/styleguide/#perspective) 😅

fuddl:
outch, I have already deleted the branch

fuddl:
to be continued in #187

Issue #180, opened 2020-04-18
Add brand emoji: LinkedIn


dayvista:
I am using OpenMoji for a project that involves the use of social media icons. Every icon I need in that regard is covered in the "brand" section except for the LinkedIn logo. I have an icon from an alternative source that will suffice in the meantime, but it subtly breaks the continuity of the design. I am requesting the creation of a LinkedIn logo OpenMoji. I've attached some reference photos. 

I understand that this is a bit of a selfish request. As such, I want to express my gratitude for this project even existing, as well as for anyone that considers my proposal.

![888065_logo_512x512](https://user-images.githubusercontent.com/58375683/79643083-304e1d00-816f-11ea-8eff-b42c118cae29.png)

![linkedin](https://user-images.githubusercontent.com/58375683/79643085-3512d100-816f-11ea-9605-d5114ff14aa3.png)

carlinmack:
would you like to have a shot at making it yourself? I can give you a hand if you'd like :)

carlinmack:
it was way harder than I realised and I've opened a PR #181 I need to see what my team mate thinks before I can merge

dayvista:
I'm not much of a designer but I'm open to trying if you'd point me in the right direction. Although, from your last response @carlinmack I'm assuming that you already went ahead and attempted to make it

carlinmack:
We have an alphabet, and it would've been simple enough to put the characters in a box, unfortunately it's only uppercase :( 

You can preview my design by clicking the link above and going to "files changed"

dayvista:
It looks v. good to me thus far. We'll wait and see what your team member thinks

carlinmack:
Added in 248f281c702dca0ddbc483c2a4d92da58047ffd9 :D Feel free to use any of the assets there in your project before they are officially released :)

PR #181, opened 2020-04-18
Add linkedin #180

carlinmack:


b-g:
@carlinmack Many thanks for this. Will have to quickly reach out to @dnlutz about how to deal with this. Currently the black stroke is not optimal. Stay tuned.

PS. Many thanks for managing all the PR over the weekend in such a nice way 🙏!!

dnlutz:
I think the only problem here is the line width modulation in the upper half of the n.
If you delete the modulation and use a consisten stroke width then it might look much better.

And what about increasing the overall stroke with of the two letters?
That would slightly enlarge the white inner forms – which might be good for the whole thing.

carlinmack:
I think I am using a consistent stroke width? maybe it doesn't look like it optically.

Wouldn't increasing stroke width decrease the size of the inner forms? I'll decrease the stroke width because I actually agree with more inner form

carlinmack:
thoughts now?

dnlutz:
sorry but it's hard to explain … i was referring to the width of the white interiour shapes ; )
as it's hard to tell how ro optimise i gave it a try myself and made the letters a bit bolder. the upper half of the "n" is adjusted manually. what do you think?
Bildschirmfoto 2020-04-25 um 14 38 24

carlinmack:
looks great :)

dnlutz:
thanks! letters look easy but are tricky in the details ; )
is there any way to upload the final svg without using the github desktop app?

carlinmack:
you could attach the file as a zip here (or host it anywhere like dropbox, Google drive, your personal website) and I could download it and do all the git work :)

dnlutz:
https://www.dropbox.com/s/jwqwkuys4du9j64/linkedin.svg?dl=0
Thank you! Github and me are still strangers to each other.

carlinmack:
closed in 248f281c702dca0ddbc483c2a4d92da58047ffd9

Issue #183, opened 2020-04-19
web feed icon


fuddl:
A symbol that i always felt was [missing in unicode](https://github.com/vhf/unicode-syndication-proposal). A symbol that represents a [web feed](https://en.wikipedia.org/wiki/Web_feed)

![FEED](https://user-images.githubusercontent.com/842548/79685325-3c51e180-8238-11ea-9fe7-271ee235858e.png)
[Figma Link](https://www.figma.com/file/sZZLEmdnMTOAxq0mTgO4WN/tardis-emoji?node-id=9%3A0)

What do you think?

carlinmack:
from looking at this in more detail online, it seems that the two arcs are actually quarter circles. Yours seem to be more than a quarter as they hug the inner circle. Could you change them? thanks

fuddl:
you mean like this?

![FEED](https://user-images.githubusercontent.com/842548/79690751-59001080-825c-11ea-876a-3603721afbba.png)

carlinmack:
![image](https://user-images.githubusercontent.com/6676843/79691197-1d4d5280-822c-11ea-91fd-4e2d7db31996.png)
could this gap be the same? after that it should be good for a PR 👍

fuddl:
![FEED](https://user-images.githubusercontent.com/842548/79692708-e9dbe980-8266-11ea-9357-680f50c8412b.png)

carlinmack:
😍

fuddl:
thanks 🎉

PR #185, opened 2020-04-19
adding web syndication symbol

fuddl:
refs #183


carlinmack:
this appears to be the old version with the wrong spacing 
image

fuddl:
🤦‍♂️

fuddl:
as it turns out `stroke-alignment="center"` can't be exported:
 stroke-alignment

probably because no browser supports it yet 😅

carlinmack:
I think that just refers to where the stroke itself is drawn

image

 I feel like the centering we're talking about is positional and should be supported?

fuddl:
fixed

PR #186, opened 2020-04-20
added power symbol

fuddl:
The [power symbol](http://www.unicode.org/L2/L2014/14009r-power-symbol.pdf) which not even displayed in osx yet.
23FB

I based it on the [AV symbol group](https://openmoji.org/library/#group=symbols%2Fav-symbol) because I thought they might be used together. They also share [the same block](https://en.wikipedia.org/wiki/Miscellaneous_Technical#Block)


carlinmack:
Nice! Can you make the height of the (circle + line) group 50.1px, so that it is the height of the circle in the template?

carlinmack:
also as these symbols all use the same assets could you also make the other 3 power symbols? 
image
which are codepoints 23FD, 2B58, 23FB and 23FC

fuddl:
> Can you make the height of the (circle + line) group 50.1px, so that it is the height of the circle in the template?

I actually based it on the diameter of the existing [record button](https://openmoji.org/library/#group=symbols%2Fav-symbol&emoji=23FA). But it might not make much sense :D

fuddl:
> also as these symbols all use the same assets could you also make the other 3 power symbols?

POWER SYMBOL/POWER ON-OFF SYMBOL/POWER ON SYMBOL/heavy circle/POWER SLEEP SYMBOL

@carlinmack btw I gave you edit permission in the [figma document](https://www.figma.com/file/sZZLEmdnMTOAxq0mTgO4WN/new-openmoji?node-id=12%3A52)

carlinmack:
I only ended up changing the record button, I think they all look good now

carlinmack:
also it seems that figma uses a different coordinate system as the circle I'm describing is 58px diameter in figma

fuddl:
@carlinmack please review

carlinmack:
btw you can request my review formally through github here 
image

carlinmack:
23FE doesn't depict a crescent moon, see this for an example https://commons.wikimedia.org/wiki/File:Power_sleep_black_cropped.svg

It should be filled in black like shown but with rounded corners rather than sharp corners

fuddl:
you mean rounder then this? cresent moon

carlinmack:
That's what I mean — the corners are round

fuddl:
It turns out, I can't assign reviewers. Do I need a certain permission to do that?

fuddl:
@carlinmack is there anything else I can do?

carlinmack:
sorry I'm just slow in getting to this! lots of outstanding issues to attend to :)

carlinmack:
thinking about it,  you could consider expanding the circuit symbols in the style of https://openmoji.org/library/#search=e1d7&emoji=E1D7 :)

fuddl:
> thinking about it, you could consider expanding the [circuit symbols](https://en.wikipedia.org/wiki/Electronic_symbol) in the style of https://openmoji.org/library/#search=e1d7&emoji=E1D7 :)

Since I never used these symbols before, I don't feed confident to do that. Do you have a use case for those? This should probably be discussed in a new issue.

carlinmack:
no worries! I was just trying to think of some more popular geometric shapes you could make :)

PR #187, opened 2020-04-20
tagged isometric

fuddl:
continuing #179


carlinmack:
Sorry that this has been such a process, but there's another problem. openmoji.csv is actually generated by `node helpers/generate-data-tables.js` and so this tagging will be overwritten the next time that this is run. The tags should actually be added in `enhancements-emoji-unicode-data.csv` and `extras-openmoji.csv`, and then the above command should be run to add the tags to `openmoji.csv` and `openmoji.json`.

I hope this makes sense! if you want to be done with this, I'm happy to make the changes instead :)

fuddl:
sorry, I am confused.  🧊 has no tags in `data/enhancements-emoji-unicode-data.csv`

carlinmack:
yeah, this file is the extra annotations we put on top of the [unicode tags](https://unicode.org/emoji/charts-13.0/emoji-list.html#1f9ca)

 you can just add `"isometric"` in between the two commas

PR #190, opened 2020-04-23
Better view of all emojis

carlinmack:
image

image

image

image

let me know if you want to change the name of the file - `generate-library.js` or the output `library.html`. Run with `node helpers/generate-library.js`


b-g:
Super! Very nice! I think I would go for something like this ...

- `generate-catalog-debug.js` obsolete -> remove
- `generate-index-list-html.js` (now called generate-catalog.js): generates index-list.html (now called index.html)
- `generate-index-html.js` (now called generate-library.js): generates index.html

OK?

carlinmack:
sounds great! 👍

b-g:
@carlinmack Great! Nice to have this super overview always online and ready: https://hfg-gmuend.github.io/openmoji/! Thanks!

PS. But why pngs and not svgs?

carlinmack:
I started with svgs but the performance was really bad, switching between black/color is very slow etc

you can check it out and if it seems feasible you can make the switch

b-g:
OK! I see ... nevermind :)

Issue #193, opened 2020-04-25
⏺ vs. ☉


fuddl:
Currently the `⏺`-Character looks very similar the [astronomical symbol for sun: `☉`](https://en.wikipedia.org/wiki/Astronomical_symbols#Symbols_for_the_Sun_and_Moon)

There is also overlap with various other shapes `⦿`, `⦾`, `⊙`, `⊚`, `◉` but with `⏺` the solution might be easy:
[why don't we make the inner dot red?](https://www.google.com/search?q=record+button&client=firefox-b-d&source=lnms&tbm=isch&sa=X&ved=2ahUKEwigmKOemYPpAhUBzaYKHeeHBpAQ_AUoAXoECA8QAw&biw=1280&bih=726)

![record-red](https://user-images.githubusercontent.com/842548/80275794-e2dc2d80-86e3-11ea-9c49-b19823b0189b.png)

carlinmack:
sure! please keep the black inner dot and add the red dot in the `color-foreground` group

PR #195, opened 2020-04-25
record button circle should have the deeper red

fuddl:
as per #193


carlinmack:
Hi please use #d22f27 or #ea5a47

fuddl:
I chose the darker one.

PR #166, opened 2020-04-05
Improve consistency of prohibited signs #164

carlinmack:

PR #201, opened 2020-04-30
Fixing some glitches

loominade:
glitches


carlinmack:
Yes great work! lots of these SVGs are messy, see #166

Issue #196, opened 2020-04-26
cleaning up figma files


fuddl:
![figma-export](https://user-images.githubusercontent.com/842548/80307052-12179b00-87c7-11ea-9393-9b21ed80c000.png)

When using these settings in the current version of [Figma](https://www.figma.com/) (_Figma Desktop App release 85_) the following updates need to be performed:

```js
const prettyfyFigmaSVG = (srcFilePath, destFilePath,) => {
  const dom = new JSDOM(fs.readFileSync(srcFilePath, 'utf8'));
  const doc = dom.window.document;
  let modified = false;

  // Problem: figma will add a `height` attribute to the `<svg>` element

  const svgWithHeight = doc.querySelector('svg[height]');
  if (svgWithHeight) {
    svgWithHeight.removeAttribute('height');
    modified = true;
  }

  // Problem: figma will add a `width` attribute to the `<svg>` element.

  const svgWithWidth = doc.querySelector('svg[width]');
  if (svgWithWidth) {
    svgWithWidth.removeAttribute('width');
    modified = true;
  }

  // Problem: figma will add a `fill="none"` attribute to the svg element. all child 
  // elements will implicitly be set to `fill="none"`. we instead want svg's default
  // behavior.

  const svgWithFillNone = doc.querySelector('svg[fill="none"]');
  if (svgWithFillNone) {
    svgWithFillNone.removeAttribute('fill');
    const elementsWithoutFill = svgWithFillNone.querySelectorAll('circle:not([fill]), ellipse:not([fill]), path:not([fill]), polygon:not([fill]), polyline:not([fill]), rect:not([fill])')
    elementsWithoutFill.forEach(el => {
      el.setAttribute('fill', 'none');
    });

    modified = true;
  }

  // Problem: figma will automatically add ids based on the type of element. These ids 
  // might end with a number.

  const figmaDefaultIds = ['Rectangle', 'Vector', 'Line', 'Arrow', 'Ellipse', 'Star', 'Polygon', 'Union', 'Subtract', 'Intersect', 'Exclude', 'Group'];
  const elementsWithId = doc.querySelectorAll('[id^="' + figmaDefaultIds.join('"], [id^="') + '"]');
  elementsWithId.forEach(el => {
    el.removeAttribute('id');
    modified = true;
  });

  // Problem: figma will create a root element `<g>` with an id that is equal to the art 
  // board name

  const rootGroup = doc.querySelector('svg > g[id]:only-child');
[...]

PR #202, opened 2020-05-01
initial commit figma cleanup script

fuddl:
refs #196 , @b-g

## How to test

1. open this file in figma:  
   [https://www.figma.com/file/sZZLEmdnMTOAxq0mTgO4WN/new-openmoji?node-id=169%3A0](new-openmoji.figma)
2. export the frame using these settings:  
  figma-export
3. choose any directory in `./src`
4. run `node helpers/prettyfy-figma-svg.js`

## Expect

1. running `node helpers/prettyfy-figma-svg.js` should remove the `.figma` interfix.
interfix removed
2. The following changes should have been applied to the svg:
    1. remove `fill="none"` attribute from the svg element. Elements without fill should have a `fill="none"` attribute.
    2. no element should have a figma default id (like `Rectangle 4`,  `Vector 7` etc…)
    3. remove root `<g>` element

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -63,6 +63,7 @@ Fix:
 Tech:
   * Blazing fast new build system by Kris Kowal @kriskowal
   * OpenMoji Black/Colorfont scale up by 1.3 to match emoji size of other vendors
+  * New cleanup script for figma svg export
 
 --------------------------------------------------------------------------------
 
--- a/helpers/README.md
+++ b/helpers/README.md
@@ -31,6 +31,7 @@ This folder `helpers/` Contains various helper scripts e.g. to export to .png an
 - `generate-index-html.js` generate overview grid index.html
 - `generate-index-list-html.js` generate overview list index-list.html
 - `import-svg-to-src-folder.js` import new OpenMojis to src folder
+- `prettyfy-figma-svg.js` prettfy svg files exported by [Figma](https://www.figma.com/)
 
 
 
--- a/helpers/prettyfy-figma-svg.js
+++ b/helpers/prettyfy-figma-svg.js
@@ -1,0 +1,71 @@
+const fs = require('fs');
+const path = require('path');
+const _ = require('lodash');
+const chroma = require('chroma-js');
+const KDTree = require('kd-tree-javascript').kdTree;
+const JSDOM = require('jsdom').JSDOM;
+
+
+const writeSvg = (filePath, data) => {
+  fs.writeFileSync(filePath, data);
+}
+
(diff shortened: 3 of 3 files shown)


fuddl:
should I also call svgo here? and add the `id="emoji"`?

b-g:
@fuddl Many thanks for this! Looks good! But don't get:

> should I also call svgo here? and add the id="emoji"?

Should we add a rule in svgo that adds a missing `<svg  id="emoji"/>` ?

b-g:
Many thanks for this! Great! Gave it a try. 

- https://openmoji-tester.glitch.me/ still gives me errors: broken color, wrong stroke and missing  a missing `<svg id="emoji"/>`. I guess the first ones are irrelevant ... but the missing id should be addressed with a new svgo rule.
- (And more a question not an issue: I've noticed a lot of `fill-rule="evenodd"`in the svg markup. What is this?)

Hence could you:
1.
- Add a rule to ensure `<svg id="emoji"/>` in svgo
- Add the a description of the script to https://github.com/hfg-gmuend/openmoji/blob/master/helpers/README.md
- Mention it in the changelog

2.
- Add it to https://openmoji.org/styleguide/#submission via a PR on https://github.com/hfg-gmuend/openmoji-website

fuddl:
>     * https://openmoji-tester.glitch.me/ still gives me errors: broken color, wrong stroke and missing  a missing `<svg id="emoji"/>`. I guess the first ones are irrelevant ... but the missing id should be addressed with a new svgo rule.

Yes I was asking if this script should handle that error or if a more generic script should be used.

>     * (And more a question not an issue: I've noticed a lot of `fill-rule="evenodd"`in the svg markup. What is this?)

this only makes sense in self-intersecting and self-containing paths. I would not be able to tell, if this is necessary  for a given geometry. [documentation of `fill-rule`](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/fill-rule)

>     * And an idea could also to call `node_modules/.bin/svgo src/extras-openmoji/brand/tmp.svg --config helpers/beautify-svg.yml` on the final files (after the rename)

That's what I wanted to as in my first comment.

> Hence could you:
>     * Add a rule to ensure `<svg id="emoji"/>` in svgo
>     * Add the a description of the script to https://github.com/hfg-gmuend/openmoji/blob/master/helpers/README.md
>     * Mention it in the changelog
>     * Add it to https://openmoji.org/styleguide/#submission via a PR on https://github.com/hfg-gmuend/openmoji-website

aright

b-g:
@fuddl Super! Many thanks for this! Merged!

PR #204, opened 2020-05-01
Data access API for OpenMoji npm module

### LINKED ISSUE #191: require() openmoji?
I was looking at the [npm repository](https://www.npmjs.com/package/openmoji) for this project and there's a button to ["Try on RunKit"](https://npm.runkit.com/openmoji). This opens an environment with the dependency already installed for devs to try out. It says that it can't automatically import anything as there is no index.js. Should we make this file? What would this enable? @kriskowal would an openmoji dependency help your project at all?

axelpale:
Fix #191

### DIFF
--- a/API.md
+++ b/API.md
@@ -1,0 +1,84 @@
+OpenMoji API
+========
+
+Access OpenMoji emoji images, their metadata, and colors from your Node.js application.
+
+## Install
+
+Install with [npm](https://www.npmjs.com/package/openmoji):
+
+    $ npm install openmoji
+
+## Usage
+
+    > const openmoji = require('openmoji')
+    > const moji = openmoji.openmojis[0]
+    > moji.emoji
+    '😀'
+    > moji.hexcode
+    '1F600'
+    > moji.openmoji_tags
+    'smile, happy'
+    > moji.openmoji_images.color.svg
+    '/path/to/your/local/openmoji/color/svg/1F600.svg'
+    > openmoji.color_palette.colors[23]
+    '#186648'
+    > openmoji.color_palette.skintones.hair[4]
+    '#000000'
+
+## API
+
+### openmoji.openmojis
+
+An array of openmoji emoji data. A openmoji emoji datum has the following structure:
+
+```
+{
+  emoji: '😀',
+  hexcode: '1F600',
+  group: 'smileys-emotion',
+  subgroups: 'face-smiling',
+  annotation: 'grinning face',
+  tags: 'face, grin',
+  openmoji_tags: 'smile, happy',
+  openmoji_author: 'Emily Jäger',
+  openmoji_date: '2018-04-18',
+  skintone: '',
+  skintone_combination: '',
+  skintone_base_emoji: '',
+  skintone_base_hexcode: '',
+  unicode: 1,
+  order: 1,
+  openmoji_images: {
+    black: {
+      svg: '/path/to/your/local/openmoji/black/svg/1F600.svg'
+    },
+    color: {
+      svg: '/path/to/your/local/openmoji/color/svg/1F600.svg'
+    }
+  }
+}
+```
+
+The paths under `openmoji_images` are absolute paths to image files located inside the `openmoji` package.
+
+### openmoji.color_palette
+
+Colors and skintones allowed in the emojis.
+
+    {
+      colors: <an array of color strings>
+      skintones: {
+        fitzpatric: <an array of color strings>,
+        shadow: <an array of color strings>,
+        hair: <an array of color strings>
+      }
+    }
+
+where an array of color strings is structurally similar to:
+
+    [ '#debb90', '#c19a65', '#a57939', '#6a462f', '#352318' ]
+
+### openmoji.version
+
+The current package version string.
--- a/index.js
+++ b/index.js
@@ -1,0 +1,28 @@
+const version = require('./package.json').version
+const openmojisRaw = require('./data/openmoji.json')
+const colorPalette = require('./data/color-palette.json')
+// For path joining
+const path = require('path')
+// Path to SVG directories
+// TODO expose svg files under src/:group/:subgroups?
+const baseBlackDir = path.join(__dirname, 'black')
+const baseColorDir = path.join(__dirname, 'color')
+
+// Enrich the emoji data objects.
+const openmojisWithPaths = openmojisRaw.map(moji => {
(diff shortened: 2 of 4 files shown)


API.md
```diff
@@ -0,0 +1,84 @@
...
+
+Install with [npm](https://www.npmjs.com/package/openmoji):
+
+    $ npm install openmoji
+
+## Usage
+
+    > const openmoji = require('openmoji')
+    > const moji = openmoji.openmojis[0]
+    > moji.emoji
```
b-g, line 16: I would rather go with `emoji`, `openmoji`, `om`, `o` and not with `moji`.

index.js
```diff
@@ -0,0 +1,35 @@
+// npm module entry
```
b-g, line 1: Add semicolons ... I know this is a religious thing ... but we are currently using them everywhere :) Please;

index.js
```diff
@@ -0,0 +1,35 @@
...
+//
+// Provides OpenMoji data in a web developer friendly fashion.
+//
+const version = require('./package.json').version
+const openmojisRaw = require('./data/openmoji.json')
+const colorPalette = require('./data/color-palette.json')
+// For path joining
+const path = require('path')
+// Path to SVG directories
+// TODO expose svg files under src/:group/:subgroups?
```
b-g, line 11: ?

test/module-index.js
```diff
@@ -0,0 +1,41 @@
+const fs = require('fs');
```
b-g, line 1: plz remove unused imports

test/module-index.js
```diff
@@ -0,0 +1,41 @@
...
+const argv = require('optimist').demand('openmoji-data-json').argv;
+const openmojiDataJson = argv['openmoji-data-json'];
+const openmojis = require(openmojiDataJson);
+
+const openmoji = require('../index');
+
+describe('Data integrity of index.js exports', () => {
+
+  describe('The paths in openmojis', () => {
+    openmoji.openmojis.forEach(moji => {
```
b-g, line 17: emoji

test/module-index.js
```diff
@@ -0,0 +1,41 @@
...
+
+const argv = require('optimist').demand('openmoji-data-json').argv;
+const openmojiDataJson = argv['openmoji-data-json'];
+const openmojis = require(openmojiDataJson);
+
+const openmoji = require('../index');
+
+describe('Data integrity of index.js exports', () => {
+
+  describe('The paths in openmojis', () => {
```
b-g, line 16: We try normally to write it as a question e.g. "Paths to SVG production files existing?" see the other tests.

index.js
axelpale: Yes, I tried but apparently missed some by habit 😅 I will add them soon. The project does not seem to have any JS linter configured. Personally, I like to use (Javascript Standard Style](https://standardjs.com/) and its linter but its rules are quite different than what has been followed in the openmoji code base.

index.js
axelpale: There seems to be a duplicate set of coloured SVG emoji files released within the npm package. I assume the ones under `src/` go through some kind of optimisation before ending up in `color/svg/`. The ones under `src/` could be left out from the npm package by appending the dir to `.npmignore` file. That would reduce npm package size that already is quite big.

test/module-index.js
axelpale: Ah, good call. A linter would be helpful to catch these. If I may suggest as a side note, [eslint](https://eslint.org/) would be the way to go.

test/module-index.js
axelpale: I see. I tried to follow the intention of mocha's `describe` and `it` methods. The `describe` block presents a module, feature or a subfeature and `it` blocks define how the feature should behave. For example see https://mochajs.org/#getting-started

However, I think it is ok to describe a question. Feels a bit unusual but I guess it works well in the output. I will convert the strings into questions.

b-g: Yes + thanks. But it is helpful as we show the same "question" in the testing results in https://openmoji-tester.glitch.me/ ... and the realized that this is more readable for all audiences.

index.js
b-g: Yes. Aware of it. We decided to ship with the src files ... but happy to discuss this. I don't think anyone is using the src files via npm install. I'm happy to do either way. It would save us ca. 7 mb of file size footprint.

@carlinmack What do you think? Should we exclude the src folder from the npm package y/n?

API.md
axelpale: I aimed to provide distinction between the emoji the character and the emoji the datum. What is the singular form of "apples"? Apple. What is the singular form of "openmojis"? Openmoji. But we cannot say "openmoji" because the package has already reserved that name. The package could be assigned into `om` or something else like `const o = require('openmoji')` but I think it is a good habit to allow developers write `const openmoji = require('openmoji')`. It reduces the number of names to remember.

I think `om` as an emoji data unit sounds fun. I would choose it.

    const openmoji = require('openmoji')
    const om = openmoji.openmojis[0]
    om.emoji

Small things but these will travel far once established. Thoughts?

b-g: Very happy with om. Many thanks for sharing your rationale behind it! I feel this entire OpenMoji things is all about details :)

index.js
carlinmack: I'm not sure, wasn't there another discussion about whether people should generate assets or not? 

I think, if we can, we should bundle the color/black svgs so that developers can start using the package out of the box rather than having to generate assets which may cause problems. We should also have a note somewhere that 72x72 will improve performance but should only be used at sizes under X.

Ryuno-Ki: What about moving it to a new issue, so it won't block this PR, then?

axelpale:
@b-g See the new resolving commits. Regarding the duplicate SVG files, I did not provide paths for the source SVGs but did not add them to .npmignore either. Removing them might be a breaking change. I think the suggestion by @Ryuno-Ki to postpone the issue sounds good.

I believe the future will reveal us more data about how developers would like to use the OpenMoji emojis programmatically. The issue #191 and work done in [openmoji-sprites](https://axelpale.github.io/openmoji-sprites/) hopefully provide a good start.

b-g:
@axelpale Great! + Many thanks! Merged!

carlinmack:
Just for the record, the original "runkit" use case in #191 won't be fixed until next release, but that's fine :)

PR #214, opened 2020-05-09
Standardise subfamily names

mavit:
It’s [strongly recommended][1] that `subfamily` be one of the four basic
styles, “regular”, “italic”, “bold”, “bold italic”.  Hence, “Color” and
“Black” should be part of the `family`, not the `subfamily`.

[1]: https://docs.microsoft.com/en-us/typography/opentype/spec/name#name-ids

### DIFF
--- a/font/scfbuild-black.yml
+++ b/font/scfbuild-black.yml
@@ -11,9 +11,9 @@ width_space: 561
 table_name:
   version: 12.2.0
   copyright: OpenMoji © HfG Schwäbisch Gmünd CC-BY-SA-4.0
-  family: OpenMoji
+  family: OpenMoji Black
   # Subfamily is also called Style or Weight. Often set to: Regular
-  subfamily: Black
+  subfamily: Regular
   unique_id: OpenMoji Black openmoji.org
   full_name: OpenMoji Black
   # No spaces in PostScript Names
--- a/font/scfbuild-black.yml
+++ b/font/scfbuild-black.yml
@@ -11,9 +11,9 @@ width_space: 561
 table_name:
   version: 12.2.0
   copyright: OpenMoji © HfG Schwäbisch Gmünd CC-BY-SA-4.0
-  family: OpenMoji Black
+  family: OpenMoji
   # Subfamily is also called Style or Weight. Often set to: Regular
-  subfamily: Regular
+  subfamily: Black
   unique_id: OpenMoji Black openmoji.org
   full_name: OpenMoji Black
   # No spaces in PostScript Names
@@ -29,3 +29,7 @@ table_name:
   url_designer: http://openmoji.org/about.html
   license: Creative Commons Share Alike 4.0
   url_license: https://creativecommons.org/licenses/by-sa/4.0/
+  typographic_family: OpenMoji
+  typographic_subfamily: Black
+  wws_family: OpenMoji Black
+  wws_subfamily: Regular
--- a/font/scfbuild-color.yml
+++ b/font/scfbuild-color.yml
@@ -11,9 +11,9 @@ width_space: 561
 table_name:
   version: 12.2.0
   copyright: OpenMoji © HfG Schwäbisch Gmünd CC-BY-SA-4.0
-  family: OpenMoji Color
+  family: OpenMoji
   # Subfamily is also called Style or Weight. Often set to: Regular
-  subfamily: Regular
+  subfamily: Color
   unique_id: OpenMoji Color openmoji.org
   full_name: OpenMoji Color
   # No spaces in PostScript Names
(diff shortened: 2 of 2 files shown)


carlinmack:
This makes a lot of sense, thanks!

b-g:
@mavit Many thanks for the feedback! But I have to say ... the way I understand it, "subfamily" is a synonym for "style". Hence for me "family: OpenMoji" and "style: Color" makes more sense that what your are suggesting. I hope you understand.

> It’s strongly recommended that subfamily be one of the four basic
styles, “regular”, “italic”, “bold”, “bold italic”. Hence, “Color” and
“Black” should be part of the family, not the subfamily.

This is news to me. What about:
Screen Shot 2020-05-12 at 12 02 54

In addition ... changing this would be a lot of effort. We would have to test all mayor browsers and apps (we had a lot of problems with this in Adobe CC) and refactor a few things. Hence a lot of pain for very little gain.

Long story short I would really vote not do this. Will close it ... but I'm happy to dicuss this if you have more good arguments for it. I hope this is fine for you. All the best!

carlinmack:
I originally agreed with the issue as I can't use OpenMoji in Word/InkScape/anything that's not a browser and I thought that this may be the issue

The source above also has:

> For extended typographic families that includes fonts other than the four basic styles (regular, italic, bold, bold italic), it is strongly recommended that name IDs 16 and 17 be used in fonts to create an extended, typographic grouping.
> 
> Within an extended typographic family that includes fonts beyond regular, bold, italic, or bold italic, distinctions are made in the Font Family name, so that fonts appear to be in separate families. In some cases, this may lead to specifying a Subfamily name of “Regular” for a font that might not otherwise be considered a regular font. For example, the Arial Black font has a Font Family name of “Arial Black” and a Subfamily name of “Regular”. Note that, in such cases, name IDs 16 and 17 should also be included, using a shared value for name ID 16 that reflects the full typographic family, and values for name ID 17 that appropriately reflect the actual design variant of each font.

So I guess we should just make sure that we are using color/black in groups 16/17? I'm not familiar with how we would do this

mavit:
> I'm happy to dicuss this if you have more good arguments for it.

From my point of view, the argument in favour is that standard font naming is a [prerequisite](https://docs.fedoraproject.org/en-US/packaging-guidelines/FontsPolicy/#_fontconfig) for fonts being included in Fedora.  This isn't a hard requirement, since we can remap the names within Fedora, but of course it's better to do this kind of work upstream if possible so that other people don't have the same bother.

> So I guess we should just make sure that we are using color/black in groups 16/17?

If 1 and 2 remain non-standard, we'd need 21 and 22 instead.

> I'm not familiar with how we would do this

I've raised [a pull request for `scfbuild`](https://github.com/13rac1/scfbuild/pull/21) to make this possible.

carlinmack:
Can all 18 styles of Inter be used on Fedora? https://rsms.me/inter/#weights

mavit:
It's not packaged for Fedora, but I just grabbed it and apparently so:

image

Why do you ask?

carlinmack:
because if Fedora supports those styles, hopefully it can support our options :)

mavit:
I pushed 2c9e687 to the branch associated with this pull request, but GitHub didn't notice, presumably because the pull request is closed.

carlinmack:
btw fyi we want to move away from OpenMoji Black/Color to OpenMoji (Lines|Strokes|Outline)/Color as Black is a pretty standard font weight that might confuse people. You're welcome to make a suggestion, but we'll probably only merge once we've decided on a name

Cc @b-g @dnlutz

b-g:
@mavit Many thanks for this! As is the scfbuild repo is no longer responsive (I did a PR in Feb 2020 and never got a response) ... Could you push your PR to our fork https://github.com/b-g/scfbuild? 

Happy to give it a try then. Is the mapping of the names happening in the background automatically or will I have to adjust / add some properties in our `scfbuild-black.yml` and `scfbuild-color.yml`

mavit:
> @mavit Many thanks for this! As is the scfbuild repo is no longer responsive (I did a PR in Feb 2020 and never got a response) ... Could you push your PR to our fork https://github.com/b-g/scfbuild?

[Done](https://github.com/b-g/scfbuild/pull/1).

> Happy to give it a try then. Is the mapping of the names happening in the background automatically or will I have to adjust / add some properties in our `scfbuild-black.yml` and `scfbuild-color.yml`

You'll have to add some properties, and I've updated this pull request to do that.

b-g:
@mavit Thanks!! Exported the font files with the new docker image https://github.com/hfg-gmuend/openmoji/commit/0f052eea437d41caab54f9e4420e1301b00abb10. Let me know if this better now.

mavit:
Looks good to me!

PR #212, opened 2020-05-08
adding monochrome flags and browncoat flag

fuddl:
as proposed in #211 

Monochrome flags and browncoat flag and nail+gear

PR #219, opened 2020-05-12
Generic flags: continued

fuddl:
@b-g follow up of #212 

> Could you do a new PR so that it appears again in the repo? Sorry for the back an forth ... but basically your PR broke our new build system and I have to figure out now why. Sorry to get entangled into this.

here we are

> @fuddl Quick question: I just released that you have choosen **extras-unicode** for all your flags. Hence is every Unicode Hexcode you've used in extras-unicode 100% official? Please confirm ... we otherwise have to move them to **extras-openmoji/flags**.

to my knowledge they are all codepoints outside the PUA. But I can check again.


b-g:
Thanks! Will investigate why this is breaking our build system ...

> to my knowledge they are all codepoints outside the PUA. But I can check again.

I think we should move it to `extras-openmoji/flags`. Currently we just have **single-hex-code** standard unicode items in `extras-unicode` and no combinations of them. What do you think?

fuddl:
I have no optinion about which folder should be used. My understanding was `openmoji-extras` is exclusively meant for single codepoints within the PUA. If it's not, I'm fine with it

carlinmack:
Yeah I think we should have the non-official codepoints in extras-openmoji. However as the [texas flag has cross compatibility](https://emojipedia.org/flag-for-texas-ustx/) I'd like to keep it in extras-unicode, does that make sense?

If we have consensus we can update CONTRIBUTING.md to advise where to put emoji

b-g:
@carlinmack Yes! I would keep everything with cross compatibility e.g. Texas and Hacker-Cat in extras-unicode and anything else in extras-opnemoji.

b-g:
@fuddl @carlinmack Hopefully finally merged and refactored. That was tricky! :)

fuddl:
Thank you and sorry for causing so much trouble

b-g:
No worries! :) Definitely an improvement ...

PR #223, opened 2020-05-16
Make mask more realistic

carlinmack:
with corona and all I feel like it should be slightly more detailed/realistic
image

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -72,6 +72,7 @@ Fix:
   * size of Computer Mouse
   * alignment of Trackball
   * detail in Couch and Lamp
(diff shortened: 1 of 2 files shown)


b-g:
Good call + nice!

Maybe: The lower part of the mask could be a bit simpler ... so that there is more padding between the mask bottom and the circle?

2020-05-18 at 08 50

b-g:
@carlinmack Many thanks! Looks super! Merged! (Should I change the authorship to you now?)

carlinmack:
Yeah I think enough has been changed to say so :)

PR #228, opened 2020-05-22
fixed attribution

fuddl:
since I created that from scratch in #218

I think i added it to `data/openmoji.csv` back then.


b-g:
@fuddl Sorry! We might lost along the way. Merged.

PR #229, opened 2020-05-22
fixed wrong stacking for white background

fuddl:
i just figured the white background shouldn't go on top:


b-g:
@fuddl thanks for spotting and the fix! Merged.

PR #239, opened 2020-06-07
3 symbols redesigned and 3 new unicode symbols

carlinmack:
I saw ℗ used in spotify and found a small list of symbols
image

Asking for review as we're the only platform putting TM in a circle https://emojipedia.org/trade-mark/ If you think we should lose the circle I'll change the service mark symbol too


b-g:
Many thanks! Great!

> If you think we should lose the circle I'll change the service mark symbol too

Yes ... I would get rid of the circle. Please

Will be busy the next days ... please merge on your own :) Thanks!

PR #198, opened 2020-04-26
added subdivision flags with bears

fuddl:
Berlin to California

PR #248, opened 2020-07-09
add subdivision flag de-by, Bavaria

Boldewyn:
From the commit message:

> TODO: b/w generates an empty frame. The result should be either solid black lozenges or, which would be heraldically better, vertically striped ones.
> 
> The official blue is #008DC9. The emoji uses the closest approximation from the OpenMoji color palette, #61B2E4. The tester app shows all tests green.

I’ve tried to align this submission closely to [the one adding the Berlin flag](https://github.com/hfg-gmuend/openmoji/pull/198).

The source for the official color specification is https://www.km.bayern.de/download/10702_anlage_4a_manual_cd.pdf .

Re: b/w: Is there a possibility to have a hidden element in the `line` layer, that can be unveiled by the `helpers/lib/export-svg-black.js` script? Or would it be better, if I provided a dedicated black-white emoji version? The California flag shares the same problem, i.e., its b/w representation is a simple white rectangle.


b-g:
Hi @Boldewyn, Many thanks +  Nice! 

> b/w: Is there a possibility to have a hidden element in the line layer, that can be unveiled by the helpers/lib/export-svg-black.js script? Or would it be better, if I provided a dedicated black-white emoji version? The California flag shares the same problem, i.e., its b/w representation is a simple white rectangle.

Yes! It should be `line-supplement`, see https://openmoji.org/styleguide/#special_layers ... please update the PR if you have time. Cheers!

Boldewyn:
Done. I’ve added an `line-supplement` layer with solid black diamonds below the `color` layer now.

b-g:
Great! Merged! Might take while until we have a new release. Many thanks!

PR #250, opened 2020-07-19
Installation instructions for Fedora.

mavit:


b-g:
@mavit Happy to merge this! Just ... do you think the name of the package `hfg-gmuend-openmoji-fonts` is ideal? The prefix quite underpins the name of my university "HfG Schwäbisch Gmünd", which I personally find a bit too much ... especially as more and more contributions are coming from outside of HfG ... and as the name of the package will be around forever ... and this might become more and more a totally independent open source project. Hence may I ask, why not just `openmoji-fonts`?

mavit:
I arrived at this name by following [Fedora's guidelines on package naming](https://fedoraproject.org/wiki/Packaging:FontsPolicy#Naming).  If you think the project should be attributed to OpenMoji.org rather than to the university, or if you think that the university will not be releasing any more fonts in the future, then we can apply to rename the package.

b-g:
Hi @mavit, 

I see, many thanks for the explanation. It seems that this is really a special case. As the other fonts e.g. `google-droid-serif-fonts` strictly follow the same pattern, I think it makes sense to stick with it. Thanks!

Merged!

Issue #230, opened 2020-05-23
Idea: Black Hole Emoji


fuddl:
![BLACK HOLE](https://user-images.githubusercontent.com/842548/82724900-8ca9dc80-9cd9-11ea-948f-c778ca367537.png) ![BLACK HOLE BLACK](https://user-images.githubusercontent.com/842548/82724901-8d427300-9cd9-11ea-8b95-d8de86de70d8.png)

Based on a [NASA Visualization ](https://www.nasa.gov/feature/goddard/2019/nasa-visualization-shows-a-black-hole-s-warped-world) which is pretty close to a [popular depiction in the film _Interstellar_](https://www.wired.com/2014/10/astrophysics-interstellar-black-hole/).
![BH_Accretion_Disk_Sim_360_Continuous](https://user-images.githubusercontent.com/842548/82725001-44d78500-9cda-11ea-9c93-de66852a6881.gif)

Could be mapped on `🕳️`+`⬛`

carlinmack:
wow that's very cool and closely matches the gif, nice! I saw this recently https://twitter.com/matthen2/status/1263678557697396738

carlinmack:
also randomly someone rated all the "black hole" emojis and we won hahah https://twitter.com/PhilipLynchGW/status/1236633891814936578

but anyway yes I think it's good and if @b-g likes it we'll merge :)

b-g:
@fuddl Super! Just wondering whether the ring could be a bit more "visually impossible" to commemorate the black hole better? I'm thinking of an Escher triangle or a Möbius strip replacing your current ring (which could also be read as ring of Jupiter) ... but not sure whether this is feasible :)

![6b3ea2e8a1eced5a76052081b23c1c67](https://user-images.githubusercontent.com/480224/82823316-a3059300-9ea7-11ea-90d9-af96a0952442.jpg)

fuddl:
the is nothing _visually impossible_ about this depiction of a black hole. just the part of the accretion disc which is _behind_ the black hole appears warped around it's horizon. Can you further demonstrate what you propose?

axelpale:
Nice! Also for me, the ring resembles the ring of saturn quite much. Maybe the ring should be connected to the event horizon like in the gif instead of being visually separate. To give an idea what I mean, see my poorly drawn sketch below.

![sketch-blackhole](https://user-images.githubusercontent.com/992081/82892801-ca686880-9f58-11ea-8e16-75a648630284.png)

[...]

PR #252, opened 2020-07-28
added black hole emoji

fuddl:
as per #230: I's been awhile but i hope everything is in order


b-g:
@fuddl Great + many thanks! Could you `npm run generate-data-tables` to generate an openmoji.json (and commit it) so that the test can run over everything including the new black hole?

fuddl:
sorry, I thought I did

b-g:
Thanks + merged!

PR #254, opened 2020-07-30
Openmoji in JavaFX UI -> Update README.md

pavlobu:
Add link in README.md to JavaFXEmojiTextFlow library that helps to use Openmoji in JavaFX UI based applications


b-g:
Many thanks for this! We will have a look! As we are not very java savvy this might take some time. Sorry! :)

Also we currently don't give credits directly in the distribution channels list. Hence could you remove "by @pavlobu", it might otherwise feel not fair to the other authors of community distribution channels ...

pavlobu:
Hi @b-g ! 

Thank you for taking a look at it.

I've removed "by @pavlobu" 

Hope it's all good to merge. I'll be here if something is wrong.

Thanks!

b-g:
Many thanks! Merged! 👍  Will try to give it a go ...

Issue #150, opened 2020-03-28
Change the flags!


OMGShay:
### If you want to know what flag emoji characters you need to change, please see below:

- [x] Laos: The middle stripe must be blue whule the two thin ones need to be red.
- [x] Turkey: The star must be flipped sideways.
- [x] Kazakhstan: Add a yellow line near the sleeve of the flag and add a bird underneath the sun.
- [x] Maldives: Make the border thicker.
- [x] Jersey: Add the emblem.
- [x] New Caledonia: Red stripe should be in the middle and the green stripe should be at the bottom.
- [x] Niger: The circle should be a bit smaller and lines need to be thicker.
- [x] Antarctica: The blue background needs to be lighter.
- [x] Mayotte: Add the word "MAYOTTE" in red above the coat of arms.
- [x] Hong Kong SAR China: Add spots in the petals of the flower.
- [x] Aruba: The 4 pointed star requires a white outline.
- [x] Iraq: The stars should be dropped and replaced by the Arabic writing.
- [x] Mozambique: Add a yellow star.
- [x] Namibia: The circle should be a sun to look like the actual Namibia flag.
- [x] New Zealand: Stars need white outlines.
- [x] Antigua and Barbuda: The sun needs to be bigger.
- [ ] Albania: Redesign central element completely
- [ ] British Indian Ocean Territory and Diego Garcia: Add the emblem, and position it near the other vertical side of the flag.

carlinmack:
Wow!!! thanks so much, I'll get on this!

carlinmack:
https://github.com/hfg-gmuend/openmoji/compare/3d28cf3377bfc8f7c8b9bc5a810d69dc14498273...7b1578527e3751ec4fa6abf72d3a752a1a589568

b-g:
Hi @OMGShay, many thanks for spotting! This is great and actionable feedback! Keep in mind that our color palette and the level of detail is limited e.g. text on the flag is not really possible.

carlinmack:
I think the text actually works even at 72x72
![image](https://user-images.githubusercontent.com/6676843/77931000-94b94300-7279-11ea-8d75-5f6cb03a44cb.png)

carlinmack:
One restriction however is the white outline around the aruba star, can't find a way to make it look good so I've improved the background colour instead

b-g:
[...]

PR #263, opened 2020-09-12
fix(🇳🇿): white border around stars

jtothegriff:
Update the New Zealand flag, 1F1F3-1F1FF, to include the white border around the stars, per issue #150.


carlinmack:
thanks!

PR #261, opened 2020-08-24
Add subdivision flag es-ct, Catalonia

jmontane:
I followed #248 (flag de-by, Bavaria). I hope everything is right with this PR.

PR #268, opened 2020-10-11
Add subdivision flag es-as, Asturies

mikelglez:
I followed #261 (flag es-ct, Catalonia). I hope everything is right with this PR.


b-g:
Hi @mikelglez! 

Many thanks for this! I fixed a few small formal things https://github.com/hfg-gmuend/openmoji/pull/268/commits/f73352ff3682f7b1cfe0d7d12af71d058f9d9e26 ... however I feels the cross has a lot of details. Is there any way to simplify it a tiny bit to push it more in general OpenMoji style?

Screen Shot 2020-10-12 at 17 41 35

cc @dnlutz Any ideas?

mikelglez:
I tried to simplify the cross a bit more, but It's not a simply cross is an specific one https://ast.wikipedia.org/wiki/Cruz_de_la_Victoria so it's difficult to recognize it in a simpler version. Also I removed the "Α" and "ω" that should have the official flag, frequently you can find this flag without this letter so I think that's not a problem.

dnlutz:
i tried to simplify the flag a bit more to make it look more openmoji-like.
what about this one?
asturias

mikelglez:
It's what I tried to explain the central circle and the proportions are very important, your version could be any cross, not the victory cross that represents the old Kingdom of Asturias... that's the problem, the flag represent a real existing cross...

dnlutz:
we can add the circle if that's a vital part. but anyhow too many details do not fit into the overall openmoji concept. our emoji are much simpler than reality. look how we simplified the spanish flag: https://openmoji.org/library/#search=spain

mikelglez:
Ok but could this parts of the cross (into the red circle) be a little smaller, to look the proportions more similar to the original?
asturies

dnlutz:
what about this one?
asturias_02

mikelglez:
Cool! I think is recognizable enough

b-g:
👍 @dnlutz please provide the svg file ... will take care of the rest then.

b-g:
Merged! Many thanks @mikelglez 🙏

mikelglez:
Thanks to you!

Issue #258, opened 2020-08-14
what would an openmoji stick figure look like?


fuddl:
There are a few [stick figures in unicode](https://www.unicode.org/charts/PDF/U1FB00.pdf):

![stick figures](https://user-images.githubusercontent.com/842548/90270111-fad5e980-de59-11ea-902e-c10fe241a58e.png)
![stick figures](https://user-images.githubusercontent.com/842548/90270105-f90c2600-de59-11ea-88c1-b24538617321.png)

I don't think they would look like the full body _person_ emoji:
![](https://openmoji.org/data/color/svg/1F9CD.svg)

What do you think a stick figure should look like?

dnlutz:
"symbols for legacy computing" is a really cool new unicode block ; )
the stick figures could look like this?
<img width="513" alt="stick_figure" src="https://user-images.githubusercontent.com/38292227/95301254-a3b02c00-0880-11eb-9320-3100d59a120e.png">

loominade:
how about having it like a _skeleton_ version of the regular people?

![skeleton](https://user-images.githubusercontent.com/11677380/95309577-9e0c1380-088b-11eb-91e2-9376c3fc1b56.png)

dnlutz:
also an interesting option!
however further away from the unicode stick figure template.
the dress looks quite tight ; )

loominade:
I suppose it doesn't have to look exactly like the one in the unicode chart.

loominade:
I think this dress looks more comfortable:

![more comfortable](https://user-images.githubusercontent.com/11677380/95316642-6190e580-0894-11eb-8986-fdd7dd9bc35f.png)

b-g:
Hi @fuddl, hope this ping finds you all well! This is just a quick heads up ... there will be soon an OpenMoji v13.0 release. Would be lovely to have the stick figures in it too (+ sidenote: did you ever contribute the nice Firefox logo?). Many thanks for all the nice contributions!! 🙏 Best, Benedikt

fuddl:
For completeness I wanted the dressed version to be able to do the same as the naked one. But those would be openmoji extra.

   ↘                          | Stick figure… | …with dress
---------------------|--------------|--------------
[...]

PR #272, opened 2020-11-04
stick figures

fuddl:
for all your legacy computing needs:

unicode stick figures

see #258


b-g:
@fuddl Great! Many thanks for the cool "legacy" stick figure people! :) Merged!

PR #284, opened 2020-11-28
Add an EditorConfig config

mavit:
[EditorConfig](https://editorconfig.org/) is a standard for configuring indentation styles (amongst other things).  It is supported by many different text editors.

The config file added by this commit will tell supporting editors to use two spaces for indentation (except in HTML files, where four spaces will be used).  This matches the existing practice of the OpenMoji project.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,9 +1,9 @@
 root = true
 
-[*.{js,json,sh,svg}]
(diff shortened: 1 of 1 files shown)


b-g:
Good call! Many thanks for this! Makes totally sense! Merged!

PR #293, opened 2021-02-11
Add Québec flag

dblanchette:
I followed #268.

The fleur-de-lis was copied from the existing emoji ⚜ (removed border, made it white).

Let me know if there is something to change/improve!


b-g:
Hi @dblanchette, Très Bien! Many thanks for this! The svg files has a few formal issues, please validate the src files against https://openmoji-tester.glitch.me/ and fix. Cheers!

@dnlutz, could you give feedback in terms of level of details of the ⚜ element?

2021-02-11 at 13 43

2021-02-11 at 13 44

dnlutz:
even if it hurts … please simplify the symbols very much!
look what we did to the spanish flag ; )
https://github.com/hfg-gmuend/openmoji/pull/293#issuecomment-777431142

dblanchette:
@b-g Thanks, I'll check when I have some time. For check n. 1 in the `npm install and tests` step, should I change the blue to be one in the OpenMoji color palette (#61B2E4 would be the closest) even if it would not be the official color of the flag?

@dnlutz It does hurt, but I understand it does not make sense to keep all this detail in emoji size, so I will make an attempt :)

b-g:
@dblanchette 

> Thanks, I'll check when I have some time. For check n. 1 in the npm install and tests step, should I change the blue to be one in the OpenMoji color palette (#61B2E4 would be the closest) even if it would not be the official color of the flag?

Yes, please only the color in the OpenMoji https://github.com/hfg-gmuend/openmoji/blob/master/data/color-palette.json. Thanks 🙏

dblanchette:
As for the simplification, what would you think of this? (Also fixed the colors)

flag_quebec_attempt_4
flag_quebec_attempt_4_618
Original for your convenience: https://fr.wikipedia.org/wiki/Drapeau_du_Qu%C3%A9bec#/media/Fichier:Flag_of_Quebec.svg

I'm not a designer by any mean (software developer), so I'm struggling a bit.

dblanchette:
I was told the fleur-de-lises look like cacti, so here is another proposal:

flag_quebec_attempt_5_72
flag_quebec_attempt_5_618

dblanchette:
@b-g All issues fixed now. I'm still not 100% sure about the design, but it's the best I can do with very few strokes.

dnlutz:
i like the few strokes very much!

b-g:
Many thanks for all the fixes! Almost ready to go ... just for the black-only-emoji has white fills currently. Please remove. The shapes should be "cut out" so that the background becomes visible. 

2021-02-12 at 15 39

dblanchette:
Should be all good now!

b-g:
I've exported the production files with `npm run generate` but now the black-only emoji is gone. Everything image and svg is deducted from the src files automatically.

Hence you have to use the a special layer named `line-supplement` and add it to the src file, see https://openmoji.org/library/#group=flags&emoji=1F3F4-200D-2620-FE0F.

dblanchette:
Unfortunately I cannot run `npm run generate` easily on my Windows machine, but hopefully this new version is ok :)

b-g:
@dblanchette Great! Merged! Many thanks for the nice flag of Québec! Cheers! (I've added you to the changelog, made a small fix ... see commits)

PR #298, opened 2021-03-08
Minor design changes to the Brunei flag.

samhocevar:
Hello,

I made some tests on the Brunei flag and I would like to suggest these design changes. The main problem I see is that the original flag dons a crescent, which is a religious symbol that many other flags have, and it seems to be the only one that somehow got lost in the simplification. Here is a summary of what I changed:
 - bend the vertical side lines to better suggest hands
 - break the circle apart to make the lower part look more like a crescent (which it is)
 - bend the top horizontal line to make it look less like a cross (it is an umbrella)

I am not a skilled designer and I don’t feel strongly about this particular design, it’s mostly here to illustrate the point.

For reference, here is the actual flag and a before/after comparison:


dnlutz:
all your improvements sound very reasonable … and it also looks better now ; )
thank you very much! we're happy to merge it ; )

b-g:
Many thanks! Merged!

PR #299, opened 2021-03-08
Fix seams in the Qatar flag and simplify polygon.

samhocevar:
Hello,

Opening the Qatar flag (`1F1F6-1F1E6.svg`) in Chrome shows small rendering artefacts caused by the leftmost white rectangle and the middle triangles touching without overlapping (first screenshot).

Opening it in Inkscape also shows that the triangles are built with twice as many vertices as necessary (second screenshot).

To fix these two issues, I suggest merging all white shapes into a single polygon. You can see that the resulting SVG is much simpler, since the aggregate polygon is made of 20 points (instead of a combined 56 previously, not counting the white rectangle).

image
image


b-g:
Many thanks for the improvement! Merged! :)

PR #300, opened 2021-03-08
Tweak Malvinas flag

samhocevar:
There is an almost unnoticeable issue with the emblem on the Malvinas flag, where the inner wavy stroke creates a small bulge on the edge of the emblem. This does not happen with all renderers and depends on the display resolution, so it’s really, really minor, but here it is in Chrome:

image image

It happens because the tip of the stroke (highlighted in green) coincides with the edge of the emblem:

image

So I suggest trimming the edges of these strokes:

image


b-g:
Many thanks! Merged!

PR #301, opened 2021-03-08
Fix South Sudan flag star orientation.

samhocevar:
Contrary to all the other flags I know, the star in the [flag of South Sudan](https://en.wikipedia.org/wiki/Flag_of_South_Sudan) points left instead of up. Unfortunately I could not find any information on why it is so.


b-g:
Many thanks! Merged!

PR #302, opened 2021-03-08
Fix the Tajikistan flag: it has 7 stars, not 8.

samhocevar:
This PR fixes the number of stars in the central emblem, and also ensures it is centred exactly at `x=36`. Here is a before/after image for reference:

image


b-g:
Many thanks! Merged!

PR #303, opened 2021-03-08
Fix East Timor flag star orientation.

samhocevar:
Note: this PR also tweaks the star position so that its barycenter lies at `y=36` (it was slightly offset at `36.503` previously).

Here is a before/after picture with the orientation highlighted:

image

The Constitution of the Democratic Republic of East Timor indicates that the star should point to the top left hand corner. I found a [construction sheet](https://www.fotw.info/flags/tl%27uno.html) illustrating that:

image


b-g:
Many thanks! Merged!

Issue #304, opened 2021-03-15
What flags are welcome?


samhocevar:
Because of specific needs I have drawn these few flags and I am wondering whether they’d be welcome as contributions.

![image](https://user-images.githubusercontent.com/245089/111116911-0069ee00-8567-11eb-9d54-1fa9bb49ab83.png)

They are:

 - [Senyera](https://en.wikipedia.org/wiki/Senyera), regional flag for Catalonia ([U+1F3F4󠁥 U+E0065󠁳 U+E0073󠁣 U+E0063󠁴 U+E0074󠁿 U+E007F](https://emojipedia.org/flag-for-catalonia-esct))
 - [Ikurrina](https://en.wikipedia.org/wiki/Ikurri%C3%B1a), regional flag for Basque Country ([U+1F3F4󠁥 U+E0065󠁳 U+E0073󠁰 U+E0070󠁶 U+E0076󠁿 U+E007F](https://emojipedia.org/flag-for-basque-country-espv/))
 - [Gwenn-ha-du](https://en.wikipedia.org/wiki/Flag_of_Brittany), regional flag for Bretagne ([U+1F3F4󠁦 U+E0066󠁲 U+E0072󠁢 U+E0062󠁲 U+E0072󠁥 U+E0065󠁿 U+E007F](https://emojipedia.org/flag-for-bretagne-frbre/))
 - [Red and black flag](https://en.wikipedia.org/wiki/Communist_symbolism#Red_and_black_flag), used for communist and anarchist symbolism since the 1800s (no official Unicode sequence so I use U+1F3F4 U+FE0F U+200D U+1F170 U+FE0F as suggested [here](https://github.com/Crissov/unicode-proposals/issues/275))

Is there a general policy on flag additions?

b-g:
Hi @samhocevar, 

many thanks for the proposals! IMO they are all welcome cc @dnlutz. 

Just note: We already have Catalonia https://openmoji.org/library/#search=Catalonia :) ... And in generals terms we outlined the OpenMoji policy in our Code of Conduct, see https://github.com/hfg-gmuend/openmoji/blob/master/CODE_OF_CONDUCT.md#contribution-guidelines-for-non-unicode-emojis.

samhocevar:
@b-g ah, thanks! I missed Catalonia because there was a typo in the Unicode sequence (fixed in #305).

PR #306, opened 2021-03-15
Add Basque, Bretagne, and anarchist flags

samhocevar:
These flags were originally discussed in #304. I have removed the Catalonia flag from the list because it was already present.


b-g:
Many thanks! But please update from master ... there are conflicting files.

samhocevar:
OK! I have merged master and resolved the conflicts (`npm run generate-data-tables` was enough). Is that okay, or do you prefer rebases instead of merges?

b-g:
Many thanks! Merged! No need to rebase ... we are not that professional :)

PR #315, opened 2021-05-09
adding jar emoji recommended for unicode 14

fuddl:
see [Recommendations for Emoji, Unicode14.0, page 5](https://www.unicode.org/L2/L2020/20242r2-emoji-recs-14.pdf#page=5)


b-g:
@fuddl Many thanks for this! Just we are currently not able to merge v14 emojis ... we don't have the metadata yet as emojibase which we rely on is not yet supporting v14 https://github.com/milesj/emojibase/issues/102.

Comments:
- please fill the emoji
- at least for me it feels quite flat for a jar ... @dnlutz any more feedback?

Screen Shot 2021-05-10 at 10 42 37

fuddl:
> please fill the emoji

It is intended to be glass ➝ transparent

Other openmoji that represent glass are also transparent. Aren't they?

carlinmack:
We decided on white fills in this thread https://github.com/hfg-gmuend/openmoji/pull/218 however there are many openmojis that have no fill for glass :/ unsure what to suggest

b-g:
I would vote to fill "glass" white. We otherwise will run into the next problem once we tackle the OpenMoji on black #31 issue.

@dnlutz would be nice to document "glass = white" under colors in the styleguide.

dnlutz:
yes, i would also opt for the glass is white solution.
in future versions we should fill all emoji completely to get good results on dark backgrounds.

b-g:
@fuddl we are rebooting! :) Summer is over! Any news on this? See also #344 for the overall progress of Unicode 14.

dnlutz:
thank you @fuddl and sorry for being a bit late to the party …

i like the color variants but maybe we get to many openmoji when we include them all?
or can we implement them as zwj sequences with color modifiers (similar to skin tone modifiers)? @b-g 

for the design i would recommend to try the following:
jar

maybe the glass should be a bit less taller and the label a bit smaller?

dnlutz:
yes … i like it!
if you want it to be more jar-like then you could make the overall glas shape a bit more squary again?
what do you think: do we need a label or can we save that?

Issue #339, opened 2021-08-15
Some black elements in SVG sources are being rendered with current font colour in TTF font


kevincox:
It appears that some strokes on the emoji use the font colour. This looks bad when the font colour is not black as expected.

Example:

```html
<span style=color:black>😃</span>
<span style=color:white>😃</span>
<span style=color:blue>😃</span>
<span style=color:green>😃</span>
```

![image](https://user-images.githubusercontent.com/494012/129491567-47235576-4747-4d27-b410-db9f36770fbf.png)

Expected behaviour:

Most emoji should not incorporate the default font color into the design.

b-g:
Closed as discussed in #340 Sorry!

kevincox:
Sorry why is this closed? This is a bug with the generated TTF files and is still an issue even if you don't want to edit the source SVG.

kevincox:
The `openmoji.css` only affects users who are using a browser and using that style. It won't help system-font usage or other cases.

Well you have a copy of the built font file in your repo so it seems that you are taking some responsibility for the generation. The argument would make sense if the project was just the SVG files but it seems like it is also maintaining a build process for fonts using pinned versions of dependencies so it is unclear how it could be fixed anywhere else. If you think it is a bug with scfbuild I can also open a bug upstream, but it seems like we would still want to keep this bug open to update the build process and checked in font file to use the newer version.

b-g:
Re-opened: https://github.com/hfg-gmuend/openmoji/pull/340#issuecomment-936074516

b-g:
@kevincox ok. Can you think of a more descriptive name for the issue? e.g.

Problems with SVG default color in font files?

kevincox:
I'm not sure this is completely accurate either because my fix didn't completely solve the problem. How about "Some black elements in SVG sources are being rendered with current font colour in TTF font?"

mesaprotector:
I am guessing that this issue is why most of the smiley emojis are unreadable for me in Mozilla software (Firefox and Thunderbird) due to the seemingly light-gray outlines; in anything else, they look fine. 

I've had Openmoji as my default emoji font for nearly a year, but still have no idea how to resolve this.

PR #260, opened 2020-08-21
Build COLR TTFs with nanoemoji

mavit:
This is done via the container at https://gitlab.com/mavit/nanoemoji-container.

For me, the resulting COLR TTFs work in Firefox and in the Gnome desktop on Fedora 32 ~~(although they display rather small)~~.

Caveats:
- It may be premature to replace `scfbuild` for SVG builds.  If we decide to go ahead without `scfbuild`, then we’ll want to remove the remaining files related to it.  If we decide to keep `scfbuild`, then we’ll want to revert those parts of this patch.  In either case we’ll want to update the readme to link to the new outputs.
- I’ve tested this with Podman, not Docker.

Relates to #93.

PR #340, opened 2021-08-16
Remove use of default colours from emoji.

kevincox:
Remove use of default colours from emoji, this caused unexpected display differences. A common symptom was eyes that matched the text colour, this resulted in poor results when the text colour isn't black. This sets the fill and stroke colour to black for any elements where they weren't specified and it was likely expected.

This change was generated by running the following commands to add stroke and fill attributes to elements which didn't have them:

```sh
cd src/

comby -jobs 32 '<:[e~ellipse|circle|rect] :[r]>' '<:[e] fill="#000" :[r]>' .svg -in-place
comby -jobs 32 '<path fill="#000" :[a]stroke=:[b]>' '<path :[a]stroke=:[b]>' .svg -in-place
comby -jobs 32 '<:[e] fill="#000" :[a]fill=:[b]>' '<:[e] :[a]fill=:[b]>' .svg -in-place
```

(This script works by adding the attributes to all instances of the intended elements then removing it if it causes a duplicate.)

The results were then manually reviewed to revert emoji where it makes sense to match the text colour as well as fix up the emoji that were not handled correctly by the simple script.

Fixes a lot of https://github.com/hfg-gmuend/openmoji/issues/339


carlinmack:
Thanks a lot for this! Could you have a check if the problem presists with the font files linked in #260? We should ensure that we don't regress on this when we change the font build system

kevincox:
I can't actually find the font files linked there. However it appears that this also doesn't solve the issue. This is surprising to me as lines that I explicitly marked as black are still changing colour. There is a chance that I just didn't regen the right files but it appears that the files used for the font build were regened so IDK what I missed. There are a lot of generated files to keep track of here 😅

b-g:
Hi @kevincox! Sorry for the super slow response, but we really needed a break over the summer! Many thanks for putting effort into this! However I have to say that I don't want to merge this:

- OpenMoji tries to be agnostic about the design tools. Because of this the source files of OpenMoji are plain SVG
- To ensure that the SVG source code stays readable and follow a common format: we force a certain coding style with svgo (see `npm run pretty-src-svg`)
- but we follow always SVG standards and as there is a default color we are deliberately using it, see SVGO https://github.com/svg/svgo#built-in-plugins `removeUselessStrokeAndFill` and `removeUnknownsAndDefaults`

Now that following the SVG standard sometimes causes "caused unexpected display differences" is so true and very sad ... but we feel that patching OpenMoji with fixes so that other software (which is not SVG standard compatible) displays OpenMoji svg files "correctly" is really out of scope. We tried this at the beginning and always failed!

Hence I will close the issue now. Sorry! Please re-open if I missed something!

(Please don't think now that your contributions were not welcome!)

kevincox:
The problem is that this cases issues with the official build process. So are you saying that the build process should be fixed instead of the emoji?

b-g:
No I'm mean we should leave it as it is, because we currently follow the svg standard. 

I don't want to introduce anything new here, because we are busy with the very core part of the project (=crafting emojis). In addition it is not our "fault" if other software packages don't follow the svg standard. As we are a very small team of people we simply don't have the capacity to take care of edge cases or to produce font files which are production ready, just go through #260.

However if you would like to take care of implementing the "default colours patch" in our build process and write the tests for it. We are really happy to get your contribution for it. Really!

1. write test to ensure that every svg file in the folders black/svg and color/svg has a default color
2. add the "default colours patch" step into the buildsystem: see overview in https://github.com/hfg-gmuend/openmoji/blob/master/helpers/export-svg.sh and e.g. helpers/lib/export-svg-black.js

kevincox:
I'm having trouble understanding what you are trying to say here. I think the following summarizes your stance, let me know if I am misunderstanding.

1. You don't want to modify the source files. Black should be the default according to SVG so the source files are correct.
2. You are ok with modifying the build system. You think that is where the fix should be made.

If that is a correct understanding then it sounds like this is rightfully closed but https://github.com/hfg-gmuend/openmoji/issues/339 should be re-opened to track the fixing of this bug using a different approach.

b-g:
Many thanks for the summery!

1. ✅
2. Yes it would be a "nice to have". But I would refrain from calling it a bug fix. We can't fix other people's software. It is just a a workaround

Feel free to re-open.

b-g:
> I can't reopen that issue, I don't appear to have permissions.

Should I re-open the issue for you? Happy to discuss things until we have sorted it out and come to a shared understanding. Again we are grateful for your contribution!

Issue #347, opened 2021-09-27
QR code icon


loominade:
Just a generic icon for qr codes:

![QR code](https://user-images.githubusercontent.com/11677380/134909455-ae6225b0-6feb-4dda-bb1a-81f5dbe08a98.png)

b-g:
Hi @loominade! Many thanks for this! Looks super! Also: super neat detail that the OpenMoji for QR codes encodes also a web search for "openmoji"! 😂

Could you please provide a PR?

PR #348, opened 2021-10-08
adding qr code icon

loominade:
refs #347


b-g:
@loominade Many thanks + merged! Super nice one!

PR #354, opened 2021-10-20
add barcode icon

loominade:
so there is a qr code #348. Some application developers might also need a good ol' barcode.

barcode icon

I'd like it to be the codepoint after the qr code icon.


b-g:
Hi @loominade! Good call! Many thanks for this! Merged :)

Issue #372, opened 2022-02-06
Some SVG files have multiple elements with the same id attribute.


RobertWinslow:
There are 13 glyphs for which the SVG source files contain duplicate id attributes. 
According to the SVG specifications, the id attribute must assign a _unique_ name to each element. 
These duplicated names caused trouble for me when trying to parse the SVG files with another program.

As an example, you can look at the source for the spoon: [1F994.svg](https://github.com/hfg-gmuend/openmoji/blob/master/src/food-drink/dishware/1F944.svg)
There are two seperate path elements with the attribute `id="_xD83E__xDD44_" `

I have already fixed the issue, and will make a pull request shortly.

b-g:
Many thanks + Great! Merged with https://github.com/hfg-gmuend/openmoji/commit/0e555f6ad5d3a41a334850f72316a44529a61368

PR #373, opened 2022-02-06
Remove problematic id attributes from certain SVG elements.

RobertWinslow:
Some of the SVG files have multiple elements with the same id.
This violates the specifications for the SVG format, and caused problems when I was trying to parse the files with another program.

Comparing these files to the other SVG sources for OpenMoji, I found that these id attributes are nonstandard.
Most of the SVG files have descriptive ids for each group ("line", "color", "skin", etc.), but not for the elements within those groups.
As such, the easiest fix was to simply remove those problematic id attributes.

Additionally, each of these duplicate id attributes have a form similar to `id="_x2935__xFE0F__1_"`.
While scanning the files,  I also found two av-symbol SVG files which had this form of id, but without the duplication.
Although these two files don't violate the specifications for SVG, I have also removed the id attributes therein for the sake of consistency with the other files in the repo.

This pull request fixes issue #372

### DIFF
--- a/src/food-drink/dishware/1F944.svg
+++ b/src/food-drink/dishware/1F944.svg
@@ -7,12 +7,12 @@
     <circle cx="36" cy="36" r="29" fill="none" stroke="#00a5ff" stroke-miterlimit="10" stroke-width="0.1"/>
   </g>
   <g id="color">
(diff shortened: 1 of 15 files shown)


RobertWinslow:
PS: For my own use, I also fixed the exported vectors in the color/svg and black/svg folders. But I haven't included those changes in this pull request because the contribution guidelines suggested that contributors should only change the graphics in the src folder.

b-g:
@RobertWinslow Many thanks for the fix and spotting 👀! Merged! Will export the production files asap to reflect your fixes/improvements. 🙏 + 👍

b-g:
PS. @RobertWinslow added you contribution to the changelog. 👍

Issue #93, opened 2019-12-02
Improve/fix OpenMoji Colorfont (and Black)


b-g:
**The OpenMoji colorfont is currently just a proof of concept and only works in Firefox.** 

We hope to improve this to proper colorfont in the near future (or at least in the long run). Sorry + stay tuned and follow this thread for updates!

b-g:
https://github.com/hfg-gmuend/openmoji/issues/66#issuecomment-547885696

b-g:
This is to gather a few links.

Currently we use https://github.com/hfg-gmuend/openmoji/blob/master/font/README.md basically [SVGinOT Color Font Builder ](https://github.com/13rac1/scfbuild)a 3-4 year old and stale project. The resulting Colorfont seem just partly to support the standard. They just work in Firefox :(

Other links:
- The [google noto-emoji repo](https://github.com/googlefonts/noto-emoji) seems to have the tooling to generate PNG based colorfonts
- https://colorfonts.langustefonts.com/howto.html (lot's of good links, but resulting font is also just Firefox)

bohnacker:
To simplify the discussion: 
Which format should we support? Right now, there are four:
PNG based: Google's CBDT/CBLC tables and Apple's SBIX table
Vector based: Microsoft's overlapped glyphs (COLR/CPAL tables) and Mozilla and Adobe’s SVG (SVG table)

These formats are supported as follows at the time (source: https://pixelambacht.nl/2014/multicolor-fonts/):
<img width="830" alt="Bildschirmfoto 2019-12-03 um 19 30 53" src="https://user-images.githubusercontent.com/2828139/70079230-9e50d800-1604-11ea-843c-20e66e936bfe.png">
<img width="836" alt="Bildschirmfoto 2019-12-03 um 19 30 43" src="https://user-images.githubusercontent.com/2828139/70079246-a3ae2280-1604-11ea-86e2-9d462c7981c6.png">

Checking with https://pixelambacht.nl/chromacheck/ there also seems to be supported SBIX in Chrome.

My opinion: there's no way around SVG because it's much more powerful than COLR. I guess, sooner or later all browsers will support SVG Color Fonts. 

Our situation:
<img width="1824" alt="Bildschirmfoto 2019-12-03 um 19 23 40" src="https://user-images.githubusercontent.com/2828139/70080063-2e435180-1606-11ea-98dd-fc5dd5705e6d.png">
[...]

PR #375, opened 2022-02-14
A batch of minor SVG improvements/alterations.

RobertWinslow:
This pull request alters a number of SVG src files. 
None of the files have been altered in a way that changes the SVG's appearance. 
I have grouped the changes into three commits so that they can be individually evaluated.

### Commit 1: Several Very Minor SVG simplications.

This alters 5 files in a way that, I believe, is a direct improvement. Details are in the commit's description.

### Commit 2: Removed zero-length paths.

This alters 220 files. It removes SVG elements that, as far as I can tell, are not rendered by a SVG renderer that follows the SVG specifications. 

### Commit 3: Convert several transform attributes to matrix notation.

This alters 6 files in a way that should have only a negligible effect on how they are rendered. See [here](https://github.com/hfg-gmuend/openmoji/issues/93#issuecomment-1039483445) for the reason behind these changes.


carlinmack:
This pleases me a lot :) If possible it would be great to remove the transforms? I have briefly tried to research into this but I couldn't find anything automatic

carlinmack:
As in apply the transformations to the paths. If you're not sure about it don't worry :)

RobertWinslow:
> If possible it would be great to remove the transforms?

It's certainly possible. I was able to use [this Inkscape extension](https://github.com/Klowner/inkscape-applytransforms) to convert the files in Commit 3 above. (The elements had to be converted to paths first.)
You can see the results in [this commit here](https://github.com/RobertWinslow/openmoji/commit/21ffc0e49e820ac420c6df650b442ae52989b4f5), though I haven't triple checked to make sure it didn't break things.
And even rushing through, it's a bit slow of a process.

Inkscape seems to have some sort of command line interface, but it isn't clear to me how to use it, nor whether it can interface with Inkscape extensions.

b-g:
@RobertWinslow Many thanks for this! Great + merged!

However this will just fix existing OpenMojis, but not fix it from a pipeline point of view. There is a high chance that newly incoming files will share the same issues. Hence best would be to add the identified issue to the /test e.g. https://github.com/hfg-gmuend/openmoji/blob/master/test/file-format.js and look into our [svgo](https://github.com/svg/svgo) rules https://github.com/hfg-gmuend/openmoji/blob/master/helpers/beautify-svg.yml to have potentially these shortcoming automatically fixed.

PR #381, opened 2022-03-18
add CDN to distribution/usage methods

jjb:
see rendered markdown here: https://github.com/jjb/openmoji/tree/patch-1#downloads--distribution-channels


b-g:
Many thanks for this! Good call! Merged! But I will strip it down a bit in terms of sub bullet points / examples urls ... I hope you don't mind.

jjb:
sounds good, no problem. i was experimenting with making a cdn for it and then discovered jsdelivr while searching in old openmoji issues, was very happy to find that

Issue #342, opened 2021-09-04
The OpenMoji butterfly emoji depicts a dead butterfly


ddevault:
Living butterflies don't hold their forewings ahead of their head, and instead keep them further back - forming a shape more like a trapezoid. It's only when they're pinned up for display that the wings are set in the positions shown in the emoji. So the butterfly emoji depicts a dead butterfly. :frowning_face: 

![](https://upload.wikimedia.org/wikipedia/commons/6/63/Monarch_In_May.jpg)

![](https://upload.wikimedia.org/wikipedia/commons/7/73/Monarch_Butterfly_Danaus_plexippus_Male_2664px.jpg)

zerorabb:
This seems to be a thing in a lot of other emoji libraries.
![image](https://user-images.githubusercontent.com/87571051/146641608-84d54e9e-ad9a-4fab-aa2a-afc91dd15af7.png)

RobertWinslow:
How's the following replacement look? I tried to base the overall shape on [this image of a Large Blue](https://en.wikipedia.org/wiki/Large_blue#/media/File:Maculinea_arion_Large_Blue_Upperside_SFrance_2009-07-18.jpg).

![1F98B](https://user-images.githubusercontent.com/5784659/159984986-abcd6e83-fd6b-4fe5-9924-0702c54317bb.svg)

It passed all the checks on [the openmoji-tester](https://openmoji-tester.glitch.me/test). I'll make a pull request shortly. Let me know if it needs any stylistic changes.

ddevault:
Looks great!

b-g:
Many thanks + looks lovely alive! Plz PR.

dnlutz:
thank you @RobertWinslow the butterfly looks good for me!
the only detail i would tweak is the blue filling – 
there's a little white gap at the bottom of the body.

PR #383, opened 2022-03-25
Redesign of shrimp, butterfly, and turtle 🦐🦋🐢

RobertWinslow:
The redesign of the butterfly (1F98B) is as discussed in issue #342 

For the Turtle (1F422), I redesigned it to make it more similar to the other emoji sets.

For the Shrimp (1F990)... this is entirely subjective, but I just didn't feel like it was *shrimpy* enough. 

I already ran these three files through [the tester](https://openmoji-tester.glitch.me/test), (which currently seems to be down), and they passed the checks.


dnlutz:
dear @RobertWinslow 
thank you for your contribution! we talked about the butterfly already, i also like the new turtle and shrimp!
i only have a few minor optimization suggestions:
Bildschirmfoto 2022-03-28 um 11 27 28
– expaning the color between the gaps of the black outline would especially help the turtle (png above)
– i would use the lighter green for the hole body of the turtle and reserve the darker green for the shell
– the new shrimp looks better but why not keeping the two different colors to make the segments more visible?

b-g:
@RobertWinslow Many thanks for this! Also a tiny idea/comment: It might be nicer to have the feeler antennas of the the Shrimp (1F990) in red. You could use the special layer `color-foreground`for this, see e.g. the red frame of Bike and https://openmoji.org/styleguide/#submission

RobertWinslow:
Thanks for the feedback. I expanded the colored bits to fill out the lines. And the shrimp color changes make it more consistent with the other OpenMoji crustaceans. Good call there. 
image image

I did keep the dark green on the turtle's belly though. That's his plastron. :)

b-g:
@RobertWinslow Great!! Merged! Added you as the "designer" in data/enhancements-emoji-unicode-data.csv ... you might want to consider to submit your portrait for https://openmoji.org/about/#authors--team so that you are listed there as well.

PR #389, opened 2022-04-29
Normalize tags of backhand index pointing gestures

waldyrious:
- Remove incorrect references to "middle finger"
- Remove unexplained "white (up|down|left|right)" tags
- Sort the tags consistently
- Remove duplicated "point" tags
- Add "point (up|down|left|right)" tags


b-g:
Great! + Merged!

b-g:
Feel free to propose something in changelog.txt ... in case you were happen to make more PRs ... otherwise I will add a reference. We appreciate your contributions!

waldyrious:
When is the next release due? I have some more changes planned, which is likely going to take a few days to finalize and submit, and I'd rather add changelog entries once I'm done with them, since it may be worth grouping some of them as data cleanup. But if a release is imminent, feel free to add something to the changelog and I'll submit the rest of the changes later.

b-g:
Next release = v14.0, is still weeks away. We currently don't have the full Unicode Emoij v14 emojis designed ... hence still time for contributions :)

PR #388, opened 2022-04-28
Normalize trailing commas in data directory

waldyrious:
Just a small cleanup change: remove trailing commas within quoted lists. This was only present in a small minority of the entries, which were normalized to match the other ones.

PR #391, opened 2022-05-09
Fix the last instance of a trailing comma

waldyrious:
This is the last remaining instance of a trailing comma, which was not captured in #388.


b-g:
Many thanks! Merged!

PR #393, opened 2022-05-15
Flipping the symbol for ionising radiation to correct orientation

SpangleLabs:
I noticed that the radiation symbol in openmoji is upside down, especially when [compared to other emoji sets](https://emojipedia.org/radioactive/), and the ANSI spec of it.

This should flip it over


b-g:
Many thanks! Merged!

PR #397, opened 2022-06-09
Adding openmoji fetching function (Python and JavaScript) to FAQ  

kolibril13:
It's nice to quickly access the openmoji database with a simple function like `get_emoji("🎭")`.
I've implemented this function, once in python and once in JavaScript, and added them to the FAQ page.
I am not sure if the FAQ is the right place, but feel free to give better suggestions where these two functions could be mentioned.
Also, one question: From what url is it best to fetch the emojis?
I've used 
https://raw.githubusercontent.com/hfg-gmuend/openmoji/master/color/72x72/{emoji_code}.png
and
https://openmoji.org/data/color/svg/${emoji_code}.svg


b-g:
Hi @kolibril13, Many thanks for the useful snippets! Yes kind of hard to find a good spot for them ... but lets leave it as it is in FAQ. Merged.

PR #398, opened 2022-06-09
FAQ: Better javascript implementation to also support emojis like 👩‍⚕️

kolibril13:
I just found a problem in the JavaScript implementation:
Emojis like 👩‍⚕️ have two parts: 👩 +⚕️.
But the current JavaScript implementation takes only the first codePoint into account.
The new implementation will take care of both cases: Single character emojies and double character emojis.
```
code = [...emoji].map(e => e.codePointAt(0).toString(16)).join(`-`)
```
The python implementation works fine though.
Follow up to #397, @b-g .

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -108,16 +108,18 @@ This script can be added to any website:
 <html>
 <script>
     function get_emoji(emoji) {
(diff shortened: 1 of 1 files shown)


b-g:
Many thanks for tweaking it further! :) Merged!

Issue #404, opened 2022-07-14
Example code on FAQ is wrong / -FE0F ignored in URL


Joshix-1:
Python-Code from https://openmoji.org/faq/ 
```python
emoji_code = "-".join(f"{ord(c):x}" for c in emoji).upper()
url = f"https://raw.githubusercontent.com/hfg-gmuend/openmoji/master/color/72x72/{emoji_code}.png"
```
generates for https://emojipedia.org/emoji/%E2%80%BC/ (‼️) 203C-FE0F.
That is correct. (see the emojipedia link)

But https://raw.githubusercontent.com/hfg-gmuend/openmoji/master/color/72x72/203C-FE0F.png is not found.
The image instead is on https://raw.githubusercontent.com/hfg-gmuend/openmoji/master/color/72x72/203C.png

The -FE0F is missing from the URL.
https://emojipedia.org/variation-selector-16/ seems to state that it is mandatory for the Emoji representation.

b-g:
@Joshix-1 Many thanks for reporting!
@kolibril13 could you have a look and fix the example?

Joshix-1:
The same is the case for ⁉️ 
https://raw.githubusercontent.com/hfg-gmuend/openmoji/master/color/72x72/2049-FE0F.png doesn't work, but https://raw.githubusercontent.com/hfg-gmuend/openmoji/master/color/72x72/2049.png works.


I would say the issue is with the names of these files and not with the example code.

Joshix-1:
I used the following python script to check for more emojis with the same issue:
The python script needs the following library: https://pypi.org/project/emoji/
I ran it in a folder where I downloaded all the svgs from the 14.0 release
```python
#!/user/bin/env python3

from emoji import EMOJI_DATA
from pathlib import Path

for emoji in EMOJI_DATA:
    emoji_code = "-".join(f"{ord(c):x}" for c in emoji).upper()
    path = Path(f"{emoji_code}.svg")
    if not path.is_file():
        if emoji_code.endswith("-FE0F"):
            path2 = Path(f"{emoji_code.removesuffix('-FE0F')}.svg")
            if path2.is_file():
                print(f"{emoji}: ", path2, "exists instead of", path)
                continue
        # print(f"{emoji}: ", path, "doesn't exist")
```
I found more files with the same issue:
```
🅰️:  1F170.svg exists instead of 1F170-FE0F.svg
🅱️:  1F171.svg exists instead of 1F171-FE0F.svg
[...]

PR #405, opened 2022-07-17
FAQ: Better handeling of unicode value U+FE0F  with Python+Javascript.

kolibril13:
Don't merge yet, it's only a draft pull request.
Attempt to solve  #404.
@Joshix-1 can you have a look at this?
It's not yet a working solution, because sometimes this character is needed:

🦴 -> "1F9B4", can be found in https://raw.githubusercontent.com/hfg-gmuend/openmoji/master/color/72x72/1F9B4.png
🐿️ -> "1F43F-FE0F" , can be found in https://raw.githubusercontent.com/hfg-gmuend/openmoji/master/color/72x72/1F43F.png
👩‍⚕️ -> "1F469-200D-2695-FE0F" can be found in https://raw.githubusercontent.com/hfg-gmuend/openmoji/master/color/72x72/1F469-200D-2695-FE0F.png

Currently, the last example would break, because the FE0F should not be removed.
I think a distinction of cases is needed here. Am I right in the assumption, that all emojis that have more than two of these character sequences separated by a "-"  should not have removed the last `FE0F` ?

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -108,14 +108,18 @@ This script can be added to any website:
 <html>
 <script>
     function get_emoji(emoji) {
-        let emoji_code = [...emoji].map(e => e.codePointAt(0).toString(16)).join(`-`).toUpperCase().replace("-FE0F", "");
+        let emoji_code = [...emoji].map(e => e.codePointAt(0).toString(16).padStart(4, '0')).join(`-`).toUpperCase()
+        if (emoji.length === 3) emoji_code = emoji_code.replace("-FE0F", "");
         new_url = `https://openmoji.org/data/color/svg/${emoji_code}.svg`
         document.write(`<img src=${new_url} style="height: 80px;">`);
+        document.write(emoji_code)
     }
     get_emoji("🦴")
     get_emoji("🐿️")
+    get_emoji("5️⃣") // problem here 0035-FE0F-20E3 -> the FE0F is removed. javascript "replace" function has to be something like removesuffix
     get_emoji("👩‍⚕️")
 </script>
+
 </html>
 ```
 
--- a/FAQ.md
+++ b/FAQ.md
@@ -109,15 +109,15 @@ This script can be added to any website:
 <script>
     function get_emoji(emoji) {
         let emoji_code = [...emoji].map(e => e.codePointAt(0).toString(16).padStart(4, '0')).join(`-`).toUpperCase()
-        if (emoji.length === 3) emoji_code = emoji_code.replace("-FE0F", "");
+        if (emoji_code.length === 10) emoji_code = emoji_code.replace("-FE0F", "");
         new_url = `https://openmoji.org/data/color/svg/${emoji_code}.svg`
         document.write(`<img src=${new_url} style="height: 80px;">`);
-        document.write(emoji_code)
     }
     get_emoji("🦴")
     get_emoji("🐿️")
-    get_emoji("5️⃣") // problem here 0035-FE0F-20E3 -> the FE0F is removed. javascript "replace" function has to be something like removesuffix
+    get_emoji("5️⃣")
     get_emoji("👩‍⚕️")
+    get_emoji("🏳️")
 </script>
 
 </html>
@@ -143,19 +143,18 @@ import requests
 
 def get_emoji(emoji):
     emoji_code = "-".join(f"{ord(c):04x}" for c in emoji).upper()
-    if len(emoji) == 2:
+    if len(emoji_code) == 10:
         emoji_code = emoji_code.removesuffix("-FE0F")
     url = f"https://raw.githubusercontent.com/hfg-gmuend/openmoji/master/color/72x72/{emoji_code}.png"
     im = Image.open(requests.get(url, stream=True).raw)
     # image = np.array(im.convert("RGBA")) 
     return im
 
-
 get_emoji("🦴")
 get_emoji("🐿️")
 get_emoji("5️⃣")
 get_emoji("👩‍⚕️")
-
+get_emoji("🏳️")
 ```
 
 
--- a/FAQ.md
+++ b/FAQ.md
@@ -138,15 +138,20 @@ from PIL import Image
 import requests
 
 def get_emoji(emoji):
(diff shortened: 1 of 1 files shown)


Joshix-1:
🏳️ would break too as it is saved as 1F3F3-FE0F https://raw.githubusercontent.com/hfg-gmuend/openmoji/master/color/72x72/1F3F3-FE0F.png

If it doesn't really matter if it is a real emoji, wouldn't it be better to do it like GitHub and always remove the -FE0F?

If it gets only removed for the short sequences the following code should work for all emojis except the white flag:
```py
emoji_code = "-".join(f"{ord(c):x}" for c in emoji).upper()
if len(emoji) == 2:
    emoji_code = emoji_code.removesuffix("-FE0F")
```
```js
let emojiCode = [...emoji].map(e => e.codePointAt(0).toString(16)).join(`-`).toUpperCase();
if (emoji.length === 2) emojiCode = emojiCode.replace("-FE0F", "");
```

Joshix-1:
Another issue with the code I just noticed is, that it e.g. doesn't work with https://openmoji.org/library/emoji-0035-FE0F-20E3/
(The leading 0s are missing)
Fix for python:
```py
"-".join(f"{ord(c):04x}" for c in emoji).upper()
```
Fix for js:
```js
[...emoji].map(e => e.codePointAt(0).toString(16).padStart(4, '0')).join(`-`).toUpperCase()
```

kolibril13:
> 🏳️ would break too as it is saved as 1F3F3-FE0F https://raw.githubusercontent.com/hfg-gmuend/openmoji/master/color/72x72/1F3F3-FE0F.png

For me, it does not break, I could also find this one: https://raw.githubusercontent.com/hfg-gmuend/openmoji/master/color/72x72/1F3F3.png

> Another issue with the code I just noticed is, that it e.g. doesn't work with https://openmoji.org/library/emoji-0035-FE0F-20E3/
(The leading 0s are missing)

Thanks for noting this!
I've just written a new python script, that should now handle all cases properly, @Joshix-1, do you want to test this?

```py
from PIL import Image
import requests

def get_emoji(emoji):
    emoji_code = "-".join(f"{ord(c):04x}" for c in emoji).upper()
    print(emoji_code)
    if len(emoji) == 2:
        emoji_code = emoji_code.removesuffix("-FE0F")
    url = f"https://raw.githubusercontent.com/hfg-gmuend/openmoji/master/color/72x72/{emoji_code}.png"
    print(url)
    im = Image.open(requests.get(url, stream=True).raw)
   # image = np.array(im.convert("RGBA")) 
    return im
imgs = []
imgs += [get_emoji("🦴")] # Code: "1F9B4" > all good
imgs += [get_emoji("🐿️")] #  Code: "1F43F-FE0F" > can be found under "1F43F" so "FE0F" has to be removed
imgs += [get_emoji("🏳️")] #  Code: "1F3F3-FE0F" > can be found eighter under "1F3F3" or under "1F3F3-FE0F"  so  "FE0F" can be removed.
imgs += [get_emoji("5️⃣") ] # Problem with missing zero?  > solved with 04x
imgs += [get_emoji("👩‍⚕️")] #  Code: "1F469-200D-2695-FE0F" > Here, FE0F does not have to be removed.

#only for debugging:
import matplotlib.pyplot as plt

plt.figure(figsize=(20,10))
columns = 5
for i, image in enumerate(imgs):
    plt.subplot(int(len(imgs) / columns + 1), columns, i + 1)
    plt.imshow(image)
```
image

Joshix-1:
> For me, it does not break, I could also find this one: https://raw.githubusercontent.com/hfg-gmuend/openmoji/master/color/72x72/1F3F3.png

Yes, true (I just looked for files ending with -FE0F and didn't check if it is there without)
But it's weird that the flag has two representations and all the others don't.

> I've just written a new python script, that should now handle all cases properly

Yes that looks better. I just don't like the += with the list. I think something like the following would be better
```py
imgs = [
    get_emoji("🦴"),  # Code: "1F9B4" > all good
    get_emoji("🐿️"),  # Code: "1F43F-FE0F" > can be found under "1F43F" so "FE0F" has to be removed
    get_emoji("🏳️"),  # Code: "1F3F3-FE0F" > can be found eighter under "1F3F3" or under "1F3F3-FE0F"  so  "FE0F" can be removed.
    get_emoji("5️⃣"),  # Problem with missing zero?  > solved with 04x
    get_emoji("👩‍⚕️"),  # Code: "1F469-200D-2695-FE0F" > Here, FE0F does not have to be removed.
]
```

FAQ.md
```diff
@@ -108,14 +108,18 @@ This script can be added to any website:
...
+        let emoji_code = [...emoji].map(e => e.codePointAt(0).toString(16).padStart(4, '0')).join(`-`).toUpperCase()
+        if (emoji.length === 3) emoji_code = emoji_code.replace("-FE0F", "");
         new_url = `https://openmoji.org/data/color/svg/${emoji_code}.svg`
         document.write(`<img src=${new_url} style="height: 80px;">`);
+        document.write(emoji_code)
     }
     get_emoji("🦴")
-    get_emoji("🎭")
+    get_emoji("🐿️")
+    get_emoji("5️⃣") // problem here 0035-FE0F-20E3 -> the FE0F is removed. javascript "replace" function has to be something like removesuffix
```
kolibril13, line 119: @Joshix-1 I've now incorporated your changes. 
However, this one emoji has a problem now.
Can you have a look at this? emoji_code.replace should be something else.

kolibril13: maybe the if statement can be replaced with regex, e.g. something like this:
https://stackoverflow.com/a/5953325/6933377

b-g:
Hi @kolibril13, Hope you've had a nice summer! Sorry for the ultra late reply! It this PR ready to merge? :)

kolibril13:
Hi @b-g,
the summer was great, hope for you as well :)
I've just fixed the -FE0F issue in the javascript implementation, so the pr is ready to merge!

kolibril13:
Hi @b-g,
the summer was great, hope for you as well :)
I've just fixed the -FE0F issue in the JavaScript implementation, so the pr is ready to merge!

b-g:
Hi @kolibril13, Great! Many thanks! + Merged.

PR #406, opened 2022-07-24
Normalize quoting of fields in extras-unicode.csv

waldyrious:
- Unquote values in the `annotation` column, which are simple strings, not lists
- Quote values in the `openmoji_tags` column, which are lists, by definition

Both of these changes were applied to lines that deviated from the existing quoting pattern, so what's being done here is merely normalizing the quoting that is already applied in this file.


b-g:
Great! Many thanks for spotting! Merged.

PR #407, opened 2022-07-24
Split `annotation` and `openmoji_tags` for E24D in extras-openmoji.csv

waldyrious:
The `annotation` field was set to "return, back button", and the `openmoji_tags` was empty.

Typically the `annotation` field is not a comma-separated list, but a simple string, so it was changed to be simply "return" (which aligns with the other ui-element entries below, such as "close", "duplicate", "forward", etc.) and "back button" was moved to the `openmoji_tags` field.


b-g:
Great! Merged.

PR #411, opened 2022-11-15
Fix a few flag paths

samhocevar:
These commits have no visible effect, they only apply reasonable rounding (*e.g.* `16.9999` → `17`) and remove duplicate points in some polygons.


b-g:
Hi @samhocevar, many thanks!! merged :)

PR #421, opened 2022-12-02
Re-flip radiation symbol + other minor SVG improvements

RobertWinslow:
The first commit re-implements the change in #393. There, the change was applied to the exported color SVG instead of the source SVG. As a result, the change was reverted on export.

The next two commits are of a similar nature to #420, and shouldn't change the visual presentation of the images.

_Edit: I also noticed that Nepal's flag was completely lacking black layers, and fixed that as well. My changes shouldn't change the appearance of the color version._


b-g:
@RobertWinslow Many thanks! Merged! ❤️

PR #431, opened 2023-01-07
Add Rounded Chinese Folk Symbols. Fú Lù Shòu etc.

RobertWinslow:
This pull request adds 6 unicode characters, representing special highly-stylized rounded versions of several Chinese characters.

- 🉠, 1F260, fú, good fortune
- 🉡, 1F261, lù, prosperity
- 🉢, 1F262, shòu, longevity
- 🉣, 1F263, xǐ, happiness
- 🉤, 1F264, shuāngxǐ, double-happiness (used especially at weddings)
- 🉥, 1F265, cái, wealth

These characters aren't used in sentences, but rather as standalone symbols. Even though they are not part of the emoji standard, they feel very 'emoji-adjacent' to me. See the following for the original unicode proposal for these symbols: [Six New Symbols from Chinese Folk Religion](https://www.unicode.org/L2/L2014/14278r2-folk-symbols.pdf) It contains a good description of the symbols and their use.

fulushou combined - no border


RobertWinslow:
I've also created enclosed versions of each of these (not included in the pull request).

fulushou_compilation_outline

If you would prefer these versions of the designs, just let me know, and I can swap them in.

(In my testing, the non-enclosed versions retain legibility at significantly smaller sizes.  I do think the enclosed version of Fu looks much better than the non-enclosed version, but I've committed the non-enclosed version for consistency with the other designs.)

b-g:
@RobertWinslow Super! Very nice addition! I believe from the OpenMoji Stylegiode perspective the enclosed versions fit better overall to OpenMoji. Hence I would vote to use the enclosed ones instead. cc @dnlutz

RobertWinslow:
SRC files have been swapped for their enclosed versions.

b-g:
@RobertWinslow Many thanks!! Merged!

PR #432, opened 2023-01-11
Adjust tags of Vulcan salute emoji (U+1F596)

waldyrious:
- Remove redundant openmoji_tags that are already present in emojibase
- Add 'fingers' and 'star trek' as openmoji_tags


b-g:
Many thanks! Merged!

PR #436, opened 2023-01-18
Adjust tags of some face emojis

waldyrious:
Just a few small adjustments to make these (mostly negative) faces more findable under common terms they're used to represent.


b-g:
@waldyrious Many thanks! Merged!

PR #437, opened 2023-01-23
Font format recommendations

mavit:
There's two parts to this:

##  Don’t recommend COLRv1 for the web

As I understand it, COLRv1 offers two advantages over COLRv0:
- Gradients (which are something that Openmoji doesn’t use).
- More reuse of elements, leading to potentially smaller file sizes.

The COLRv1 TTF file is indeed smaller.  However, for some reason the WOFF2 file is not:

| Size    | Filename                         |
| ------- | -------------------------------- |
| 2436372 | OpenMoji-color-glyf_colr_0.ttf   |
|  905144 | OpenMoji-color-glyf_colr_0.woff2 |
| 2220984 | OpenMoji-color-glyf_colr_1.ttf   |
|  972648 | OpenMoji-color-glyf_colr_1.woff2 |

Hence, I don’t think there’s any reason to recommend COLRv1 WOFF2 at the current time.

## Be more generous about support for COLRv0 on desktop

COLRv0 is the built-in format for colour fonts in Windows 10, so I don’t think it’s fair to say that it has very little support.  These days, it’s also supported by many Linux applications too.


b-g:
@mavit Many thanks for pointing this out! Can't verify all this. Simply trusting the real experts :) cc @RobertWinslow Merged.

PR #448, opened 2023-05-04
Fix key emojis to not show color outside of the outlines

rosseljost:
Some parts of the color fill were showing beyond the lines (especially noticeable on dark backgrounds), slightly adjusted the paths accordingly.

Old:

New:


b-g:
@rosseljost Many thanks! Merged!

rosseljost:
I just noticed this change is not reflected in the webpage: https://openmoji.org/library/emoji-1F511/
Is there a reason for that?

b-g:
Yes. Every release is a lot of manual work :) Also we try to keep the version number in sync with Emoji Versions, ans as we still don't have all new upcoming emojis in the lib ... there won't be soon a new release. Sorry! Please be patient. (But of course you can use the master branch with all the new things.)

Issue #399, opened 2022-07-01
Emojis for communication channels: XMPP, Matrix / Element, E-Mail with OpenPGP


Lukas2112:
I suppose it might make sense to extend the OpenMoji project in Republic Communication to include XMPP, Matrix / Element and Email with OpenPGP key icons.

b-g:
Hi @Lukas2112, many thanks for the suggestions ... yes makes sense! But please be more visual and provide images of the existing logos. Not everyone is super familiar with those tools / services. Cheers.

Lukas2112:
Thanks for your feedback! Of course I can do that, @b-g 

The logo of XMPP can be found at https://upload.wikimedia.org/wikipedia/commons/9/95/XMPP_logo.svg , that of Element at https://commons.wikimedia.org/wiki/File:Element_(software)_logo.svg and that of Matrix at https://de.wikipedia.org/wiki/Datei:Matrix_logo.svg here.

PR #451, opened 2023-05-23
feat: add element icon

SethFalco:
Adds the Element logo in the OpenMoji style.

Here is the original SVG of the logo for comparison:

The shade of green is a bit shoddy, but I think it's the best we can do with the limited color palette. :thinking: 

### Related

* Partially fulfills https://github.com/hfg-gmuend/openmoji/issues/399


b-g:
Many thanks! @SethFalco merged! :)

Issue #485, opened 2023-12-23
Wrong version number in TTFs


mavit:
The version metadata in the TTF files in the 15.0 release incorrectly say that they are version 14.0.0.  This is because the version number is guessed from the latest Git tag, and of course, the new tag can't be created until after the TTFs have been generated.

![image](https://github.com/hfg-gmuend/openmoji/assets/120296/9659255f-9b42-4a23-becc-31f37a83078b)

PR #486, opened 2023-12-23
Provide a way to specify the version of TTFs

mavit:
This should be done when building TTFs for release, so that the new fonts can be committed before the Git tag is created.

Provides a way to prevent recurrences of #485.


b-g:
@mavit Many thanks for spotting and the quick fix, great! But could we not put the version string to `data/OpenMoji-Color.ttx` as well? I would find this more logically to have all font meta infos in one place. What do you think?

(will fix the font version string once back in office)

PR #493, opened 2024-03-21
White Spots

RealityRipple:
I've identified 24 emojis with white elements in the background of the glyph. In some cases, it seems more intentional and stylistic, but in all cases, it makes them feel out of place on colored or dark backgrounds.

I've split the list up into things I'm more certain about and less certain about:
 - Emojis in "Set A" have oddly shaped white elements that definitely look out of place.
 - Emojis in "Set B" seem a little more stylistic, but still look wrong on close inspection.
 - Emojis in "Set C" are more "filled-in-holes" and background white. While not visually jarring, they do seem to assume a light background.

This PR is not intended to be directly merged - it's more of a reference. Some of these emojis may be better served with a different approach than just removing the white spots entirely.


b-g:
Dear @RealityRipple, many thanks for eagle eyes! I wasn't aware to this "side effect". I'm calling it a side effect, as we are working in the background on making the openmojis finally work as well on a dark background e.g. see https://github.com/hfg-gmuend/openmoji/pull/231#issuecomment-643671423 ... hence we've iterated and tweaked all existing openmojis to dark-mode-ready = that we can automatically generate a white halo around it. You can test this by uploading an openmoji to our testing app ( https://openmoji-tester.glitch.me) and then running the `3. Check visually (for dark mode)`  test. Long story short ... we won't be fixing this, as these are necessary for the upcoming dark mode support. But if you feel energized, happy to get help on this. Al the best.

Issue #467, opened 2023-09-13
U.S. subdivision flags


maxterry:
As with other emoji sets, looks like only England, Scotland, and Wales are currently included as subdivision flags in OpenMoji:

https://openmoji.org/library/#group=flags%2Fsubdivision-flag

State and territory flags for the U.S. are defined and possible, just not implemented: https://blog.emojipedia.org/us-state-flag-emojis-now-possible

Would be cool to get them included here.

Related to https://github.com/hfg-gmuend/openmoji/issues/416

Kirylchyk:
hey @maxterry, I picked up and started to work on this enhancement. https://github.com/hfg-gmuend/openmoji/pull/489 was merged to add Washington DC flag emoji. I am aware of the following flags https://en.wikipedia.org/wiki/Category:United_States_state_flags and planning to add flags at the reasonable pace. Do you have any order preferences? Thank you

xavizardKnight:
@Kirylchyk @alexcochran Heyas! I want to start advancing with the US subdivision flags, now that the subdivision flags of Spain are done, and I've seen that both of you have contributed some recently. Have you done any more advancements on other flags? Can I continue on working on the remaining flags?

Kirylchyk:
hey @xavizardKnight that's awesome! Indeed, i'm working on some more emoji. Lets split

maxterry:
Nice to see some momentum on this! Would be great to work toward a complete set of U.S. subdivision flags eventually, no preference from my side on order/priority to get there.

xavizardKnight:
> hey @xavizardKnight that's awesome! Indeed, i'm working on some more emoji. Lets split

Yeah! If you started some, give me the list of those you haven't started yet. Or let's split the list 50/50 and start working from there, as you prefer.


> Nice to see some momentum on this! Would be great to work toward a complete set of U.S. subdivision flags eventually, no preference from my side on order/priority to get there.

We can try to aim and see if we can complete all the US subdivision flags before the next OpenMoji release; there are quite a few flags, and some of them are pretty complex, but I think we can manage to pull it off.
Don't know when it will be the next release, but probably by the end of this year.

xavizardKnight:
Also, I've realised that the recently added Colorado flag has no grid layer in its source file. Technically, it does have a grid layer, but it's hidden. This doesn't seem to affect the outputs, but it looks weird in comparison with the rest of the flags when viewing the thumbnails of the folder. Was this intentional?

[...]

PR #496, opened 2024-04-09
Readding the grid for the Colorado flag

xavizardKnight:
Quick commit for readding the grid of the Colorado flag, as debated in issue #467. No metadata changes, only the source vector file, as this is a single simple edit to an existing emoji.

PR branched off from master. Merging this PR should not close any issue.


b-g:
Many thanks for spotting this 🪲. Merged.

PR #495, opened 2024-04-09
Transferring 16 emojis, mainly Spain subdivision flags (April 2024)

xavizardKnight:
Heyas! This is the PR for transferring the next batch of 16 emojis that I've been talking in some issues lately, including Spain subdivisions and four more brand icons. This time, I've written the correct month on the branch name, that's already an improvement! xD

Included emojis in this PR:
- extras-unicode/subdivision-flag/1F3F4-E0065-E0073-E0061-E006E-E007F.svg (Andalusia ES-AN)
- extras-unicode/subdivision-flag/1F3F4-E0065-E0073-E0063-E006E-E007F.svg (Canary Islands ES-CN)
- extras-unicode/subdivision-flag/1F3F4-E0065-E0073-E0063-E0062-E007F.svg (Cantabria ES-CB)
- extras-unicode/subdivision-flag/1F3F4-E0065-E0073-E0063-E006C-E007F.svg (Castile and León ES-CL)
- extras-unicode/subdivision-flag/1F3F4-E0065-E0073-E0063-E006D-E007F.svg (Castile La Mancha ES-CM)
- extras-unicode/subdivision-flag/1F3F4-E0065-E0073-E0063-E0065-E007F.svg (Ceuta ES-CE)
- extras-unicode/subdivision-flag/1F3F4-E0065-E0073-E0065-E0078-E007F.svg (Extremadura ES-EX)
- extras-unicode/subdivision-flag/1F3F4-E0065-E0073-E0067-E0061-E007F.svg (Galicia ES-GA)
- extras-unicode/subdivision-flag/1F3F4-E0065-E0073-E0072-E0069-E007F.svg (La Rioja ES-RI)
- extras-unicode/subdivision-flag/1F3F4-E0065-E0073-E006D-E0064-E007F.svg (Madrid ES-MD)
- extras-unicode/subdivision-flag/1F3F4-E0065-E0073-E006D-E006C-E007F.svg (Melilla ES-ML)
- extras-unicode/subdivision-flag/1F3F4-E0065-E0073-E006E-E0063-E007F.svg (Navarre ES-NC)
- extras-openmoji/brand/E04C.svg (Dotnet / .NET programming framework logo)
- extras-openmoji/brand/E04D.svg (Golang programming language logo)
- extras-openmoji/brand/E04E.svg (Kotlin programming language logo)
- extras-openmoji/brand/E04F.svg (Ruby programming language logo)

Captura de pantalla a 2024-04-08 18-46-08

[...]

Issue #468, opened 2023-09-14
Emojis for programming languages


streambinder:
It'd be nice to extend the Brands category adding the most important programming language logos, such as C, Python, Rust, and so on and so forth.

b-g:
Done with https://github.com/hfg-gmuend/openmoji/commit/d9f538710ff8e60417b3bf8a0e901424172108bc. Closing.

streambinder:
Thanks for picking this one up, but the PR is missing some relatively important languages, such as Go, Python, Rust. Any plan these will get added as well?

xavizardKnight:
Heyas!
Sorry for replying after many months, I've been missing-in-action from the OpenMoji projects for a few months for personal issues.

I'm the one who made the few programming language icons that were added in the latest release (mainly the C, C++, C#, JavaScript, TypeScript, WebAssembly, SVG and Markdown). I've made these icons because they were very simple to do as I learned SVG and Inkscape, and all of them are under public domain, which makes them compatible with this project's licence.

As @b-g described in issue #475, there are no descriptions for emojis in the openmoji-extras section (where these programming icons would go in the library); thus, icons that require attribution cannot currently be added to this repo, as there would be no way to attribute them.

The quickest way to check if an icon or logo of something has a compatible licence is searching it on Wikipedia and look if it has public domain / CC0 licences.

Let's go through a list of the most popular programming languages and frameworks and see if they would be possible:

- ❌ **Ruby:** CC-BY-SA. But, see note below.
- ✅ **Go:** Public Domain.
- ❌ **Python:** GPL.
- ❌ **Rust:** CC-BY.
- ❌ **PHP:** CC-BY-SA.
- ❌ **HTML5 / CSS3 / JS / TS shields:** all of them are CC-BY.
- ✅ **Kotlin:** Public Domain.
- ❌ **Java:** Copyright.
- ✅ **.NET:** CC0. I actually have this icon created in my repo, but I did not remember to push it to here in my previous PR… xD. Will transfer it on my next PR.

Note:
Because Ruby's icon is a literal ruby (as a gemstone), an icon of Ruby could also refer to the gemstone, and the latter meaning is public domain.
The same thing happens with Emerald, a Linux window decorator program, whose icon is a green emerald tilted clockwise.
[...]

PR #498, opened 2024-04-12
Let's go: Golang

xavizardKnight:
Let's talk about the Golang icon.

I've pushed 5 versions of the Golang icon on the /src/ folder. These are just tests, and the final emoji will be moved to the correct place when finished.

The original emoji that I presented in PR #495 is `golang-smaller.svg`. `golang-bigger.svg` is the same, but bigger (following your suggestion on the original PR of making the letters bigger). `golang-blue-smaller.svg` and `golang-blue-bigger.svg` are the same as the non-blue ones, but without the black outline at the GO letters. `golang-original.svg` is using the letters of the original logo instead of the OpenMoji font. The original has a line between the G and O, but I think removing it is better in this case.

Which one do you prefer the most? You should have read and write access to this branch, so feel free to experiment how you wish.

Once we decide on a version, I'll rename it, move it to the correct place, remove the others and readd the metadata.

Related with issue #468 and PR #495.


dnlutz:
My favourite is golang-original.svg. It looks more dynamic than the upright OpenMoji letters so I would suggest to go for this one. Thank you!

b-g:
I agree with Daniel ... but @xavizardKnight what do you think?

b-g:
And many thanks for taking the time to prepare five versions! 👍 🙏

xavizardKnight:
Hi! Sorry for the super delay that I'm replying lately; I'm having a complicated year.

I agree that the version with the original lettering style looks better than the rest. Although it does look weird (at least for me) for an icon with such big letters to not use the OpenMoji font…

I'll start to tidy up this branch to prepare the original lettering style emoji to merge it soon. Will ping when everything's ready!

b-g:
Great! Hope all is good on your end! Yes ... this will be a special case :)

b-g:
@xavizardKnight That's great! Many thanks! + Merged.

PR #502, opened 2024-06-01
Update 1F3F4-E0066-E0072-E0062-E0072-E0065-E007F.svg

le-sigwal:
First & last lines in black (merged with outside border to improve rendering at small scale). x11 stoats as usual

### DIFF
--- a/src/extras-unicode/subdivision-flag/1F3F4-E0066-E0072-E0062-E0072-E0065-E007F.svg
+++ b/src/extras-unicode/subdivision-flag/1F3F4-E0066-E0072-E0062-E0072-E0065-E007F.svg
@@ -8,19 +8,23 @@
   </g>
   <g id="color">
     <rect x="5" y="17" width="62" height="38" fill="#fff"/>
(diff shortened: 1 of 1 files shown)


b-g:
Many thanks for this! However I don't want to merge this, as we don't aim to optimize small scale but rather want to have a consistent look & feel for the entire library e.g. the black frame is one of the distinctive features ... and with your changes the flag stands more out and the black frame is less distinctive. I'm sorry! I hope you see my point and I hope this not too de-motivating. All the best!

le-sigwal:
In fact the existing version looks like breton flag at small scale (within small text for example).
But at a little greater scale, it is difficult to recognize. It seems as a negative of the flag (as if the colors were swapped).
Apart from that very interesting design trade-off! 
Mik
`

b-g:
Thanks! It seems you are the expert. But happy to find a compromise e.g. a better general Breton flag (at any scale), while keeping the distinctive black frame (for the sake of the OpenMoji project). Happy to review more PRs :)

PR #520, opened 2024-10-20
Emoji 15.1 direction-specifying variations of person-activities

RealityRipple:
Flipped the relevant X values and arc rotations for these SVGs.

Note that the manual wheelchair direction was already right-facing, so those were renamed and new flipped left-facing wheelchairs were made for the older Emoji 12/12.1 entries.


b-g:
Hi @RealityRipple, many thanks for this! But running `npm run test-dev` results in quite a few errors. I won't able to accept the PR before these have been fixed. Also ideally the svg src files have not been altered for the direction change manually e.g. in Illustrator, but rather via the command line helper `node mirror-svg.js path/to/source.svg path/to/result.svg` (see folder helper). It is basically a transform attribute in each top level g element. Let me know if you want to tackle it or not, it was on my list as well. And lastly I think the author of the flipped openmojis should still be the old one, but happy to add you to the changelog for the help. Al the best

RealityRipple:
A new PR from scratch might be less messy than editing this one, then.

PR #521, opened 2024-10-22
Emoji 15.1 direction-specifying variations of person-activities (Second Draft)

RealityRipple:
Complete rewrite of PR #520

Uses `mirror-svg` instead (except on manual wheelchair, which is hard-flipped and then mirrored).

Required manually modifying `emojibase-data/en/data.json`, as these emojis are missing skin tone modifiers in their annotations for some reason.


b-g:
@RealityRipple Brilliant! Many thanks!! Merged! I guess we are Unicode 15.1 feature complete. Yay!

PR #523, opened 2024-10-23
Updated identification card

### LINKED ISSUE #522: Update the Identification Card to the new official emoji
In has an blue color. Google, Microsoft, Facebook, Toss Face and Icons8 have already updated their emoji sets to reflect the change.

nataliakeniganti:
Updated identification card emoji

Closes #522


b-g:
Many thanks for the proposal. I can't find an official Unicode source why it has to be blue. Please provide a source, otherwise I think we should stay agnostic and neutral. Not every country issues blue ID cards.

nataliakeniganti:
> Many thanks for the proposal. I can't find an official Unicode source why it has to be blue. Please provide a source, otherwise I think we should stay agnostic and neutral. Not every country issues blue ID cards.

is sadly thanks it merged!!!!!!

Issue #417, opened 2022-11-22
A red cross should not be used


maekawa-mugi:
A red cross is used in the ambulance (U+1F691) emojis.
The Geneva Convention and Japanese law limit the use of the Red Cross to the Red Cross Society in principle.

Therefore, I propose that it should be changed to a blue asterisk instead.

maekawa-mugi:
Additional Info:
Mr. Shigeta (ex. Docomo), known as the father of the emoji, spoke about the hospital emoji on Twitter.
According to his story, the Red Cross pointed out that the emoji colors have been changed.

https://twitter.com/sigekun/status/24565482257
https://twitter.com/sigekun/status/1578740490325094402

carlinmack:
I understand and in principle agree, however[ many platforms still use the red cross](https://emojipedia.org/ambulance/) and so we will delay this change unless someone contributes a pull request. I hope this is okay

b-g:
Done

Issue #385, opened 2022-04-01
Improve: gun/water pistol emoji


SpangleLabs:
Following on from discussions in #366 
The [gun/pistol emoji](https://emojipedia.org/pistol/) drew [much media attention](https://fortune.com/2018/04/25/no-more-gun-emoji/) a couple years ago after Apple moved to displaying it as a water pistol, and all other major platforms soon followed suit.
Openmoji is still depicting it as a handgun however, which makes it unsuitable as a drop-in replacement.

ikkqqstudio:
Why nobody improving this

RealityRipple:
If we're going by standards, as was implied in other issues, all instances from v13.1 onward of `emoji-test.txt` on unicode.org specify "water pistol" for `1F52B`, as of September 2020, before any of the issues on this subject were even opened.

PR #524, opened 2024-10-30
Add human arm E180, and make various improvements.

nataliakeniganti:
A src vector for the human arm symbol (codepoint E180) has been added

Additionally, several src vectors have been modified:

- Whiten cross mark button
- Fixed color black small square
- Fixed part for couple with heart

This pull request should hopefully fully address the following issues:

- #417 
- #385


b-g:
Many thanks for the improvements! Great! Just the vector path quality of the arm is not there yet, especially the lower and upper parts should look more geometric. Try to redraw those lines with fewer points, which should make them more geometric and better looking. Also the shadow seems to be slightly off around the elbow, see https://openmoji.org/styleguide/#shadow

2024-10-31 at 16 06

nataliakeniganti:
> Many thanks for the improvements! Great! Just the vector path quality of the arm is not there yet, especially the lower and upper parts should look more geometric. Try to redraw those lines with fewer points, which should make them more geometric and better looking. Also the shadow seems to be slightly off around the elbow, see https://openmoji.org/styleguide/#shadow
> 
> 2024-10-31 at 16 06
> 

Merging 10 minutes

nataliakeniganti:
it fine thanks
image

PR #570, opened 2025-09-28
easy font usage via jsdelivr.com

jjb:
Here's a recipe to use the font via jsdelivr.com

(If you like the general idea) These docs could live here, or somewhere else, or maybe be integrated into the generate-font-demo script.


b-g:
@jjb Nice one! Many thanks. Merged.

PR #574, opened 2025-10-20
A11y Improved alt text for emoji image

javice25:
Improved alt text for emoji image to make the page more accessible for screen reader users. This supports accessibility best practices (WCAG alt text rule).


Ryuno-Ki:
How so? From what I can tell it adds more clutter.

Issue #572, opened 2025-10-11
Proposal To Add Pride Flags 🏳️‍🌈


felifluid:
Heyey! i recently found out about the OpenMoji project and am using it for my digital garden. for this i designed several common pride flags. i have currently uploaded the files to a [seperate repository](https://github.com/felifluid/openmoji-prideflags). i want to suggest including them in this repository as well. 

![preview](https://github.com/felifluid/openmoji-prideflags/blob/main/preview.png?raw=true)

i am aware that in #475 @b-g stated that [OpenMoji will no longer accept any flags without unicode definition](https://github.com/hfg-gmuend/openmoji/issues/475#issuecomment-1836310967).

> We changed a while our contribution guidelines and no longer are accepting any flags without an official unicode definition. Sorry! But arguing e.g. to include a flag like Kurdistan or not is too tough. We resignated at some point as we simply couldn't moderate it. Hence we won't be able to include those (even if I may add, personally I would too). I hope you understand.

However, [the Unicode Consortium will no longer accept any proposals for flag emojis](https://blog.unicode.org/2022/03/the-past-and-future-of-flag-emoji.html). This is especially highlighted for pride flags.

> Despite being the largest emoji category with a strong association tied to identity, flags are [by far the least used](https://home.unicode.org/emoji/emoji-frequency/). (There are exceptions: usage of the [rainbow flag is above median](https://observablehq.com/@jenniferdaniel/unicode-emoji-mirror)!) That begs the question, “So, why not encode more identity flags?” Well, we have seen the same results for flags as we have seen for other emoji — a very long tail of rarely used options. They also tend to change over time! In the past six years since adding a Pride Flag to the Unicode Standard (2019) it’s already been redesigned. Many times. Identities are fluid and unstoppable which makes mapping them to a formal unchanging universal character set incompatible.

therefore it is highly unlikely that more pride flags will be added to the official unicode definiton in the forseeable future.

soo... i think OpenMoji has a special opportunity at hand here. according to its [code of contuct](https://github.com/hfg-gmuend/openmoji/blob/master/CODE_OF_CONDUCT.md) OpenMoji supports adding emojis that adress "social and critical issues e.g. Climate change" and it also includes this pledge:

> We pledge to act and interact in ways that contribute to an open, welcoming, diverse, inclusive, and healthy community.

and i think adding pride flags would be highly in support of this pledge.

[...]

PR #587, opened 2025-11-15
Submission: Add lesbian flag

elizj3:
🌸 Created by Emily Gwen (https://x.com/theemilygwen), and implemented based on OpenMoji templates by me.

✨ It's free for use (more on https://lesbianflaguse.carrd.co).

PR #588, opened 2025-11-19
Add Queer Symbols 🏳️‍🌈

felifluid:
This is the pull request for the addition of queer symbols as discussed in detail in #572. 

The new queer symbols are added to the non-unicode section `openmoji-extras` in a new subcategory `queer-symbols`. It reserves the private use area codepoints `E420–E460`.

Note that this PR copied the transgender emoji 🏳️‍⚧️ to the new codepoint `E421`. That means it can be used as a singular unicode symbol, instead of using `1F3F3-FE0F-200D-26A7-FE0F` (🏳 • VS16 • ZWJ • ⚧ • VS16) to create it. However, it think it's important to include the trans flag redundantly in these two different ways in order to ensure compatibility. 

I wasn't able to run the tests, as installing with `npm` failed on my machine (will look into that). So would be great if someone else can run those prior to merging.

Fixes: #572 and #587


b-g:
Hi @felifluid, many thanks for this! Currently it does not pass the tests ... I think it would be really easier if you could find a way to run the test locally. Otherwise it will be a lot back and forth. Once you've added/edit the meta infos in data/extras-unicode.csv, do the following:

- run `npm run generate-data-tables` to generate the files in data folder eg. openmoji.json
- then `npm run test-dev` → if green, please PR :) ... 

Also we don't have any way to have redundancy and aliases. Hence for the transgender emoji either a breaking change (moving it to E421, I would vote for that) or leaving at 1F3F3-FE0F-200D-26A7-FE0F. We have in every release breaking changes, once existing openmoji-extras become included to the most recent Unicode.

felifluid:
> Also we don't have any way to have redundancy and aliases. Hence for the transgender emoji either a breaking change (moving it to E421, I would vote for that) or leaving at 1F3F3-FE0F-200D-26A7-FE0F. We have in every release breaking changes, once existing openmoji-extras become included to the most recent Unicode.

Just for clarity: How would the current transgender emoji "🏳️‍⚧️" be displayed after this change? Would it be displayed as "🏳⚧"?

b-g:
@felifluid Yes. The current transgender emoji (the src svg file) will simply be renamed and moved to the new location in queer-symbols. Hence in future releases 1F3F3-FE0F-200D-26A7-FE0F will point to no src file and will either result displaying the "[missing glyph](https://openmoji.org/library/emoji-25A1/)" symbol or "🏳⚧" depending on the context. Basically users will have to update to the new codepoint E421.

felifluid:
@b-g All issues should have been resolved now. The old transgender emoji was removed, including the entry in [`data/enhancements-emoji-unicode-data.csv`](https://github.com/hfg-gmuend/openmoji/pull/588/commits/413012ccf32546b5b1ffaa9e9dad80783caf50b9).

b-g:
@felifluid, many thanks great! We are getting close! :) Please:

- Fix missing and mandatory grid layer for all emojis, see https://github.com/hfg-gmuend/openmoji/pull/588/commits/792ed5ee333e887aabc258b1f826f4e3b9415073. The grid layer are automatically removed on production export (I will take of this) 
- And I've made a mistake, the transgender emoji 1F3F3-FE0F-200D-26A7-FE0F is already part of Unicode. "Transgender Flag was added to [Emoji 13.0](https://emojipedia.org/emoji-13.0) in 2020." In our discussion above I was thinking the Transgender emoji would only be an openmoji-extra 😬. Hence we should not move the 1F3F3-FE0F-200D-26A7-FE0F to any new place, the code point 1F3F3-FE0F-200D-26A7-FE0F has to stay where it is, to fulfill Unicode Emoji 13.0 ... and only all the new proposed ones would go to queer-symbols. My apologies! 🙈

felifluid:
Alrights. I think this _should_ be it now? Added the grids, moved the transgender emoji back to its original codepoint and adjusted the other codepoints to fill in the gap. 

PS: This commit history could/should possibly get squashed together when merging this PR.

b-g:
@felifluid Many thanks! Yay! Merged.

Issue #594, opened 2025-12-15
Mexico subdivision flags


memburg:
I would like to get the green light to work on some [official flags in Mexico](https://en.wikipedia.org/wiki/State_flags_of_Mexico).

As of now I have Yucatan:

<img width="300" height="300" alt="Image" src="https://github.com/user-attachments/assets/554424cd-afb8-4c3e-ac77-2a9a519d24c0" />

And Queretaro:

<img width="300" height="300" alt="Image" src="https://github.com/user-attachments/assets/cf6ee4f9-9546-48eb-a4dc-e9c78400067d" />

Greetings!

b-g:
Hi @memburg, many thanks for reaching out. If the subdivision flags are official and have a representing unicode hexcode, please propose the flags. However we will only accept the flags if the abstraction is really significant and aligns nicely with the existing flags e.g. Yucatan should only have circles and no stars :)

![Image](https://github.com/user-attachments/assets/80cf6e4b-692b-4828-9836-2c76bf4a3e65)

The flag of wales is a good examples of the level of abstraction what you should be aiming for.

![Image](https://github.com/user-attachments/assets/87c3aeda-27e6-4480-8203-5de5c142d271)

memburg:
Nice @b-g 

So I think I can represent them with:

`U+1F3F4 U+E006D U+E0078 U+E0071 U+E0075 U+E0065 U+E007F` and `U+1F3F4 U+E006D U+E0078 U+E0079 U+E0075 U+E0063 U+E007F`

I'll do the proper changes so then I can fork and send a PR.

Thanks!

PR #595, opened 2025-12-19
Add: Flag of Querétaro

memburg:
Reference: https://en.wikipedia.org/wiki/Flag_of_Quer%C3%A9taro#/media/File:Flag_of_Queretaro.svg

Hello, please let me know if this is okay and if there's anything else I can add/modify.

PR #596, opened 2025-12-24
Add: flag of Yucatán and Querétaro

memburg:
Closes: #594 

This can be merged after #595 or both flag can be added at the same time.


b-g:
@memburg Many thanks for this. Looks good. Merged.

PR #627, opened 2026-03-16
Add flag of Ontario

jj-marr:
This PR adds the [flag of Ontario](https://en.wikipedia.org/wiki/Flag_of_Ontario) as a subnational flag, because that is where I live.

The original draft was created with an LLM (Claude 4.5 Sonnet in Cline) based on the [Bermuda flag](https://en.wikipedia.org/wiki/Flag_of_Bermuda) which is also a Red Ensign. AI did a terrible job on the crest, so I had to manually edit the proportions of the leafs/St George's cross. 

The flag uses the shaded versions of all colours for consistency with the Bermuda flag and because the light green/yellow look terrible in the crest. The crest is also positioned higher up, so that it looks smaller, similar to depictions of the flag of Ontario I've seen online.

### DIFF
--- a/data/extras-unicode.csv
+++ b/data/extras-unicode.csv
@@ -28,6 +28,7 @@ emoji,hexcode,group,subgroups,annotation,openmoji_tags,openmoji_author,openmoji_
 ⮻,2BBB,extras-unicode,symbol-other,overlapping white and black squares,,loominade,2020-04-22,1.1
 ⮼,2BBC,extras-unicode,symbol-other,overlapping black squares,,loominade,2020-04-22,1.1
 ⯄,2BC4,extras-unicode,symbol-other,black octagon,"equilateral polygon",loominade,2020-04-17,5
+🏴󠁣󠁡󠁯󠁮󠁿,1F3F4-E0063-E0061-E006F-E006E-E007F,extras-unicode,subdivision-flag,ontario flag,"ontario, canada, province, red ensign",JJ Marr,2026-03-16,?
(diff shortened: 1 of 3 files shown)


b-g:
Many thanks for this. Please squash the commits. I don't want to have "add questionable AI draft of flag of Ontario" in the commit history ... and deal with it after. Sorry!

Please explain the proposed codepoint: 1F3F4-E0063-E0061-E006F-E006E-E007F

Test are failing. Please check with `npm run test-dev`until everything is green.

jj-marr:
> Please explain the proposed codepoint: 1F3F4-E0063-E0061-E006F-E006E-E007F

Unicode represents subdivision flags with the waving black flag (1F3F4), then an invisible tag sequence following ISO 3166-2 that represents the country (CAON, in this case), then the cancel tag (E007F).

https://github.com/unicode-org/cldr/blob/maint/maint-41/common/validity/subdivision.xml

Unicode allows implementations to add flags for _any_ subdivision with an ISO 3166-2 code, which OpenMoji has done for other subregions like Texas. 

However, it only "recommends for interchange" England, Wales, and Scotland because they don't want to be accused of favouritism for certain regions. My goal is to add as many flags as possible, so this isn't a concern.

Other subdivisions like Ontario are just in the CLDR list which means some implementation somewhere uses it.

jj-marr:
@b-g It's squashed and passes tests.

b-g:
@jj-marr Many thanks! Merged

PR #628, opened 2026-03-17
Add flag of Bari, Italy

jj-marr:
This is the flag of the [Metropolitan City of Bari, Italy](https://en.wikipedia.org/wiki/Metropolitan_City_of_Bari).


b-g:
Many thanks! See https://github.com/hfg-gmuend/openmoji/pull/627#issuecomment-4074063183 for feedback and issues

b-g:
@jj-marr Many thanks! Merged

PR #659, opened 2026-06-15
fix link

jjb:


b-g:
@jjb many thanks for spotting and the fix! 👍 Merged.

Issue #660, opened 2026-07-13
Inconsistent stroke width of the bluesky icon


loominade:
![The Bluesky icon has a thinner stroke width compared to other icons](https://github.com/user-attachments/assets/0c499455-b5de-490e-b39c-57079cac5acc)

nataliakeniganti:
@jjb make this pull

jjb:
i think autocomplete filled in my name but you meant someone else 😅

b-g:
Fixed with #661

PR #661, opened 2026-07-13
Update stroke width in Bluesky logo

nataliakeniganti:
As per #660 

Increased stroke width for the path in the SVG.


b-g:
@nataliakeniganti Thanks for spotting and fixing. Merged

